PR #30, opened 2020-01-21
Feature/enhance create dataset

### LINKED ISSUE #27: Question on generating scores
Hello, 

I am able to run the example configuration given in the repo and the score generation to get the affinity scores for ICLR 2020. However, it is not clear to me what the config file should be for other years and conferences. For example, if I take the example configuration file and replace 2020 with 2019 it does not work. How can I figure out how to get the config for each conference for which there is data?

Thanks,

Tanner

carlosmondra:
It looks like a lot of changes, but I basically did 3 things that :

- I refactored the code in `create_dataset` so that it can be more general. It works with all the venues that I have tried, this potentially fixes issue #27.
- I also improved the way the `config.json` was loaded. Once it is loaded, the `ModelConfig`, which now iherits from `UserDict`, is instantiated.
- I replaced the `os` library with `libpath` as it seems to be more adequate from what I read and it is one of the issues that Michael created.

### DIFF
--- a/expertise/config/core.py
+++ b/expertise/config/core.py
@@ -1,14 +1,19 @@
 from collections import OrderedDict
 import json
 import random
-import os
+from pathlib import Path
 import pickle
 import pkgutil
 import expertise
 
 class ModelConfig(object):
     def __init__(self, **kwargs):
-        self._config = {}
+        if kwargs.get('config_file_path'):
+            config_file_path = Path(kwargs['config_file_path'])
+            with open(config_file_path) as file_handle:
+                self.config = json.load(file_handle)
+        elif kwargs.get('config_dict'):
+            self.config = kwargs['config_dict']
 
         # valid_model_names = expertise.available_models()
         # if not 'model' in kwargs:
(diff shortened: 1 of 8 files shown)


expertise/config/core.py
```diff
@@ -1,53 +1,35 @@
-from collections import OrderedDict
+from collections import UserDict
 import json
 import random
-import os
+from pathlib import Path
```
mohituniyal: @carlosmondra It seems pathlib works with Py3.4 and above. I think we should we should put that somewhere in the documentation that we only support Py3.4 and above.

carlosmondra: Good point! I'll say that we support only python 3.6 and above though.

expertise/create_dataset.py
```diff
@@ -6,27 +6,185 @@
...
 
 from collections import defaultdict, OrderedDict
 
-def search_with_retries(client, term, max_retries=5):
-    results = []
-    for iteration in range(max_retries):
-        try:
-            results = openreview_client.search_notes(
-                term=term, content='authors', group='all',source='forum')
+from config import ModelConfig
```
mohituniyal: This needs to be changed to `from .config import ModelConfig` to make this work.

mohituniyal:
I noticed that having an experiment_dir in the config.json is inconsequential right now as we are always creating the dataset files in root dir.

mohituniyal:
@carlosmondra The changes look good to me.

expertise/create_dataset.py
carlosmondra: Fixed!

PR #36, opened 2020-02-28
Config file allows to use emails instead of Profile IDs

carlosmondra:
This PR has some enhancements when creating the dataset. The main improvement though is that now we allow to use emails as IDs. This is because sometimes the information can come from outside OpenReview and emails can work as IDs.


expertise/create_dataset.py
```diff
@@ -46,22 +46,32 @@ def get_profile_ids(client, group_ids):
...
     :rtype: list
     """
     tilde_members = []
+    members = []
     for group_id in group_ids:
         group = openreview_client.get_group(group_id)
+
         profile_members = [member for member in group.members if '~' in member]
+        profile_search_results = openreview_client.search_profiles(emails=None, ids=profile_members, term=None)
+        if profile_search_results and type(profile_search_results) == dict:
```
mohituniyal: If you search by profile ids then the results would be a list instead of a dict. Is this being taken into account here @carlosmondra ?

mohituniyal:
@carlosmondra This change looks good to me besides the one minor comment I had.

expertise/create_dataset.py
carlosmondra: Woooo! Thank you very much for spotting this HUGE mistake. I didn't know that search_profiles returned different data types depending on the parameter. I have fixed that and removed some extra unnecessary checks.

carlosmondra: All the more reason to add tests. I think I'll need to work on those

expertise/create_dataset.py
```diff
@@ -46,22 +46,30 @@ def get_profile_ids(openreview_client, group_ids):
...
         profile_members = [member for member in group.members if '~' in member]
+        profile_search_results = openreview_client.search_profiles(emails=None, ids=profile_members, term=None)
+        tilde_members = []
+        for profile in profile_search_results:
+            preferredEmail = profile.content.get('preferredEmail') or profile.content.get('emails')[0]
+            tilde_members.append((profile.id, preferredEmail))
+        members.extend(tilde_members)
+
         email_members = [member for member in group.members if '@' in member]
+        email_set = set(email_members)
```
mohituniyal, line 65: @carlosmondra You are not using this `email_set` anywhere. So, let's remove it. Since these are group members, they will always be unique anyway.

expertise/create_dataset.py
```diff
@@ -46,22 +46,30 @@ def get_profile_ids(openreview_client, group_ids):
...
-    tilde_members = []
+    members = []
     for group_id in group_ids:
         group = openreview_client.get_group(group_id)
+
         profile_members = [member for member in group.members if '~' in member]
+        profile_search_results = openreview_client.search_profiles(emails=None, ids=profile_members, term=None)
+        tilde_members = []
+        for profile in profile_search_results:
+            preferredEmail = profile.content.get('preferredEmail') or profile.content.get('emails')[0]
```
mohituniyal, line 60: Maybe we can look for the 0th element in the `emailsConfirmed` list before relying on the `emails` list. What do you think @carlosmondra ?

carlosmondra: I agree, good point. I am checking `emailsConfirmed` before checking `emails`. In the case the user does not have an email in the Profile I fall back to the Profile ID.

expertise/create_dataset.py
```diff
@@ -46,22 +46,34 @@ def get_profile_ids(openreview_client, group_ids):
...
+            # If the user does not have emails confirmed, use the first email in the emails list
+            preferredEmail = preferredEmail or profile.content.get('emails') and len(profile.content.get('emails')) and profile.content.get('emails')[0]
+            # If the user Profile does not have an email, use its Profile ID
+            tilde_members.append((profile.id, preferredEmail or profile.id))
+        members.extend(tilde_members)
+
         email_members = [member for member in group.members if '@' in member]
         profile_search_results = openreview_client.search_profiles(emails=email_members, ids=None, term=None)
+        email_profiles = []
+        for email, profile in profile_search_results.items():
```
mohituniyal, line 72: @carlosmondra I found another issue here caused by of the `client.search_profiles` method. In the case when there are no emails in the match group members, we would be passing an empty list in the emails arg to search profiles and in that the `profile_search_results` would be returned as an empty list and the `.items()` call on it would fail.
Maybe we can code the line 70 as `profile_search_results = openreview_client.search_profiles(emails=email_members, ids=None, term=None) if email_members else {}`
What do you think?

carlosmondra: Thanks Mohit! I think I didn't do that because `search_profiles` already returns `{}` when no Profiles are found. I tested this out, just in case.
```python
>>> c.search_profiles(emails=['mondragon52369@hotmail.com'])
{}
>>> c.search_profiles(emails=['carlos@openreview.net'])
{'carlos@openreview.net': Profile(id = '~Carlos_Mondragon1',number = None,tcdate = 1559740458953,tmdate = 1580485922603,referent = None,packaging = None,invitation = '~/-/profiles',readers = ['OpenReview.net', '~Carlos_Mondragon1'],nonreaders = None,signatures = ['~Carlos_Mondragon1'],writers = ['OpenReview.net'],content = {'preferredEmail': 'carlos@openreview.net', 'gender': 'Male', 'homepage': 'https://www.linkedin.com/in/carlosmondragonchapa/', 'emails': ['carlos@openreview.net'], 'names': [{'first': 'Carlos', 'middle': '', 'last': 'Mondragon', 'preferred': True, 'username': '~Carlos_Mondragon1'}], 'history': [{'position': 'MS student', 'start': 2017, 'end': 2019, 'institution': {'domain': 'umass.edu', 'name': 'University of Massachusetts, Amherst'}}], 'relations': [], 'emailsConfirmed': ['carlos@openreview.net']},metaContent = {'preferredEmail': {'signatures': ['~Carlos_Mondragon1']}, 'gender': {'signatures': ['~Carlos_Mondragon1']}, 'homepage': {'signatures': ['~Carlos_Mondragon1']}, 'emails': [{'signatures': ['~Carlos_Mondragon1']}], 'names': [{'signatures': ['~Carlos_Mondragon1']}], 'history': [{'signatures': ['~Carlos_Mondragon1']}]},active = True,password = True,tauthor = 'OpenReview.net')}
```

mohituniyal: Sorry if I wasn't clear earlier. Try calling `c.search_profiles(emails=[])` and that would return an empty list which would then cause an exception as the code would look for `.items()` on a list. This is what will happen at line 70 if `email_members` is an empty list.

carlosmondra: Done! I pushed the changes

PR #37, opened 2020-03-05
Feature/duplicate detection

carlosmondra:
This PR adds a new functionality: Duplicate Detection.

It aims to make it easier to just provide two files: `submissions.jsonl` and `other_submissions.jsonl` to perform duplicate detection. I also updated the README.md so that the instructions for Dubplicate Detection are clear.


README.md
```diff
@@ -62,6 +62,30 @@ python -m expertise.tfidf_scores config.json
 
 The output will generate a `.csv` file with the name pattern `<config_name>-scores.csv`.
 
+## Detect Duplicates
```
mohituniyal, line 65: I know that we support duplicate detection for both within a venue and between different venues. I can see that we basically compare the data in the directories specified by `config.json` params `submissions_path` and `other_submissions_path`.
I think it would be a good idea to add a line or two here explaining how these params can be used to support both duplicate detection modes.

mohituniyal:
Tested this, looks good to me.

README.md
```diff
@@ -139,13 +162,39 @@ Here is an example:
...
+        "scores_path": "./",
+        "use_title": false,
+        "use_abstract": true,
+        "use_cuda": true,
+        "batch_size": 4,
+        "knn": 10,
+        "skip_same_id": false,
+        "skip_elmo": false,
+        "submissions_path": "./",
+        "other_submissions_path": "./"
```
mohituniyal, line 193: Oh and one more thing I realized. I had to copy over the data in the dir `experiments_dir/submissions` into a dir `experiments_dir/other_submissions` that I had to create. It would be great if I did not have to duplicate the data to be able to do duplicate detection within the conference. Maybe other_submissions can be an optional config param and the absence of which would indicate an intra-conference duplicate detection. We would also not have to run embedding on the same submissions twice thereby saving some execution time as well. What are your thoughts

carlosmondra: Yup, that's a good idea. I was being lazy haha. I'll add that option. I think that would address your comment above?

README.md
carlosmondra: I am not really sure what you mean by explaining the two modes. I can avoid having to make the user duplicate the submissions into other_submissions folder when it is intraconference duplicate detection, but I am not sure what you would like me to explain in the README.

carlosmondra:
Thanks @mohituniyal, I have addressed your comments. Let me know what you think :D

README.md
mohituniyal: By 2 modes I meant intra-conference and inter-conference.

carlosmondra: Got it! I have updated the README, let me know if you think that the explanation is clear enough.

README.md
```diff
@@ -62,6 +62,32 @@ python -m expertise.tfidf_scores config.json
 
 The output will generate a `.csv` file with the name pattern `<config_name>-scores.csv`.
 
+## Detect Duplicates
+
+There are two modes for duplicate detection: intra-conference and inter-conference. Intra-conference detection detects duplicate submissions within the same conference and inter-conference detection detects duplicate submissions with other conferences. For intra-conference detection only one `submissions` directory or one `submissions.jsonl` file is needed. For inter-conference a `submissions` directory and a `other_submissions` directory or a `submissions.jsonl` file and a `other_submissions.jsonl` file are needed. More details about these files/directories can be found at the end of this section.
```
mohituniyal, line 67: @carlosmondra how about this:

Duplicate detection can be used to find duplicates both within a venue or between 2 different venues. 
- For detecting duplicates within a venue only one `submissions` directory or one `submissions.jsonl` file is needed.
- For detecting duplicated between 2 different venues, either one `submissions` and one `other submissions` directories are needed **or** one `submissions.jsonl` and one `other_submissions.jsonl` are needed.

More details about these files/directories can be found at the end of this section.

carlosmondra: That definitely sounds better haha. I have updated the README.

PR #38, opened 2020-03-10
Feature/sparse scores

carlosmondra:
This PR creates sparse scores, by passing a value to the `config.json` called `sparse_value`. If the `sparse_value` is 100, for example, it ensures that every paper-reviewer pair gets at least 100 scores.


README.md
```diff
@@ -83,6 +83,7 @@ These parameters could be included in a separate file, like `affinity-config.jso
 - `model_params.scores_path`: This is the directory where the `.csv` file with the scores will be dumped.
 - `model_params.use_title`: Boolean that indicates whether to use the title for the affinity scores or not. If this is `true` then `model_params.use_abstract` must be `false`.
 - `model_params.use_abstract`: Boolean that indicates whether to use the abstract for the affinity scores or not. If this is `true` then `model_params.use_title` must be `false`.
+- `model_params.sparse_value`: Optional numerical value. If passed, instead of returning all the possible publication-submission combinations, only the top scores will be returned. The number of top scores will be determined by the `sparse_value`. That does not mean that the number of scores per submission will be equal to the `sparse_value`.
```
mohituniyal, line 86: Does it instead mean that `the number of scores per **publication** will be equal to the **sparse_value**`?

expertise/models/bm25/bm25.py
```diff
@@ -82,12 +89,15 @@ def all_scores_helper(self, submissions_dataset):
         csv_scores = []
         for note_id, submission in tqdm(submissions_dataset.items(), total=len(submissions_dataset)):
             reviewer_scores = self.score(submission)
+            if reviewer_scores is None:
+                continue
             for profile_id, score in reviewer_scores.items():
-                csv_line = '{note_id},{reviewer},{score}'.format(reviewer=profile_id, note_id=note_id, score=score)
+                csv_line = (note_id, profile_id, score)
+                # csv_line = '{note_id},{reviewer},{score}'.format(reviewer=profile_id, note_id=note_id, score=score)
```
mohituniyal, line 96: Do we want to keep this commented line?

carlosmondra: Nope haha

README.md
carlosmondra: Check the updated README. I hope it's clear now :D.

expertise/models/bm25/bm25.py
```diff
@@ -49,6 +51,9 @@ def set_archives_dataset(self, archives_dataset):
     def set_submissions_dataset(self, submissions_dataset):
         self.submissions_dataset = submissions_dataset
 
+    def _is_valid_field(self, obj, field):
+        return field in obj and len(obj.get(field)) > 0
```
mohituniyal, line 55: @carlosmondra how about using this instead:
```suggestion
        return field in obj and obj.get(field)
```
This avoids 2 errors:
1. If the field does not exist in obj, the `len(None)` operation would error out.
2. If the field exists but is not a type that supports len() operation, it would error out.

expertise/models/bm25/bm25.py
```diff
@@ -85,12 +92,14 @@ def all_scores_helper(self, submissions_dataset):
         csv_scores = []
         for note_id, submission in tqdm(submissions_dataset.items(), total=len(submissions_dataset)):
             reviewer_scores = self.score(submission)
+            if reviewer_scores is None:
```
mohituniyal, line 95:
```suggestion
            if not reviewer_scores:
```
Slightly more pythonic.

expertise/models/elmo/elmo.py
```diff
@@ -65,6 +66,9 @@ def set_other_submissions_dataset(self, other_submissions_dataset):
             elif self.use_title and 'title' in submission['content']:
                 self.other_sub_note_id_to_title[submission['id']] = submission['content']['title']
 
+    def _is_valid_field(self, obj, field):
```
mohituniyal, line 69: same comment as the BM25 Model class

expertise/run.py
```diff
@@ -18,23 +17,37 @@
         submissions_dataset = SubmissionsDataset(submissions_file=Path(config['dataset']['directory']).joinpath('submissions.jsonl'))
 
     if config['model'] == 'bm25':
+        from .models import bm25
```
mohituniyal, line 20: any particular reason for importing here instead of at the top of the file?

expertise/run.py
```diff
@@ -18,23 +17,37 @@
...
         bm25Model.set_archives_dataset(archives_dataset)
         bm25Model.set_submissions_dataset(submissions_dataset)
-        bm25Model.all_scores(Path(config['model_params']['scores_path']).joinpath(config['name'] + '.csv'))
+
+        if not config['model_params'].get('skip_bm25', False):
+            bm25Model.all_scores(
+                preliminary_scores_path=Path(config['model_params']['scores_path']).joinpath('preliminary_scores.pkl'),
+                scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '.csv')
+            )
+        if config['model_params'].get('sparse_value'):
```
mohituniyal, line 35: Am I correct in saying that this generates both `all_scores` and `sparse_scores`? If so, why do we need both scores?

carlosmondra: You're right, it's OK to only have one

expertise/run.py
carlosmondra: Yeah, I was reading just now that it is considered best practice to [import at the top of the file](https://stackoverflow.com/questions/1188640/good-or-bad-practice-in-python-import-in-the-middle-of-a-file). But I noticed that even if I was running BM25, the files used by ELMo were also loaded, which took time and were unnecessary. In that same thread I found [this answer](https://stackoverflow.com/a/1188693/12597574) as well. To avoid overhead is acceptable. What do you think?

mohituniyal: Right, I thought about that while commenting but I did not find either of the two classes very bulky (I might have missed something). Anyway, if you see a real difference in execution time by doing a lazy import, I think it is ok to do it.

carlosmondra: So the ELMo file is not the problem, but it seems that when I import ELMo, it also imports the AllenNLP libraries which are the ones that take time. Also, I thought it would be confusing for the user to import libraries that are unrelated to what they called. So if they use BM25, why should they see AllenNLP being loaded? haha. Maybe there's another way to structure the code that would avoid this and have the imports at the top at the same time. But I cannot think of one...

mohituniyal: How about doing this at the top of the script?
```
if config['model'] == 'bm25':
        from .models import bm25
if config['model'] == 'elmo':
        from .models import elmo
```

carlosmondra: I like that solution! I'll update the code. Thanks for suggesting it.

PR #39, opened 2020-03-27
Add support for profiles expertise and bids exported as JSON format

carlosmondra:
Adds support to retrieve assignments from Edges and Notes as well as the Expertise in the Profile.


melisabok:
Also conflicts here.

melisabok:
@carlosmondra do we still need this PR or should we close it?

carlosmondra:
There doesn't seem to be a need for this. Let's close it.

PR #42, opened 2020-04-24
Feature/normalize elmo scores

### LINKED ISSUE #41: Normalize ELMo scores
ELMo scores are usually between 0.5 and 1. It's better to normalize them to have more spread out values.

carlosmondra:
This PR adds an option to normalize the ELMo scores. It also adds a timestamp or number of publications to import from a reviewer to avoid having scores that are very close to 1.

Fixes #41


mohituniyal:
@carlosmondra The duplicate detection PR introduced this line in the readme: `Duplicate detection uses ELMo exclusively, since we normalize the scores for BM25`.
And now this PR is normalizing ELMo scores as well. Will this affect duplicate detection somehow?

carlosmondra:
> @carlosmondra The duplicate detection PR introduced this line in the readme: `Duplicate detection uses ELMo exclusively, since we normalize the scores for BM25`.
> And now this PR is normalizing ELMo scores as well. Will this affect duplicate detection somehow?

Good point, it won't affect it. But maybe we should change the README. The normalization in ELMo is optional, but should not be used during duplicate detection, just for affinity scores. In fact, if you look at the `find_duplicates` file, you'll see that the normalization param is not included even if the users sets it as `true`. I hope this answers your question

expertise/create_dataset.py
```diff
@@ -26,12 +26,71 @@ def convert_to_list(config_invitations):
...
+    # Get all publications and assign tcdate to cdate in case cdate is None. If tcdate is also None
+    # assign cdate := 0
+    unsorted_publications = []
+    for publication in publications:
+        if getattr(publication, 'cdate') is None:
+            publication.cdate = getattr(publication, 'tcdate', 0)
+        unsorted_publications.append(publication)
+
+    # If the author does not have publications, then return early
+    if len(unsorted_publications) == 0:
```
mohituniyal, line 44: Minor comment, let's write this in a more pythonic style.

```suggestion
    if not unsorted_publications:
```

expertise/create_dataset.py
```diff
@@ -26,12 +26,71 @@ def convert_to_list(config_invitations):
...
+        return unsorted_publications
+
+    dataset_params = config.get('dataset', {})
+    minimum_pub_date = dataset_params.get('minimum_pub_date') or dataset_params.get('or', {}).get('minimum_pub_date', 0)
```
mohituniyal, line 48: @carlosmondra I am not sure about this second condition in this statement. 
Is there a situation where in dataset_params we would not have 'minimum_pub_date' but a dict 'or' with the key 'minimum_pub_date' would be there?

carlosmondra: Right, I forgot to add to the README. Basically I am adding two configuration options:
- `minimum_pub_date`
- `top_recent_pubs`

Before you read the explanation, it might also help see the test case that I created for this function :D.

`minimum_pub_date` refers to the minimum publication date. Any submission that is older that this date will not be taken into account for the expertise.
`top_recent_pubs` can be a number or a percentage. If the user sets a number, like say 3, then only the 3 most recent publications are taken into consideration. The same applies for percentage. If the reviewer has less than 3 publications, then all his/her publications are taken into consideration. For percentages, even if we select 10%, if the user has 3 publications, then 10% of 3 is 0.3. Decimal values are always rounded to the next integer, so the result in this case would be 1.

Now, if both options are present, I didn't know if it was better to do an AND or an OR between the results of both constraints, so I decided to add both options. If these parameters are inside the `or` field, then I do an OR relationship if they are not, then I do an AND relationship.

README.md
```diff
@@ -103,11 +103,55 @@ Below you will find examples of possible configurations depending on the Model t
...
+    "paper_invitation": "ICLR.cc/2020/Conference/-/Blind_Submission",
+    "exclusion_inv": "ICLR.cc/2020/Conference/-/Expertise_Selection",
+    "bid_inv": "ICLR.cc/2020/Conference/-/Add_Bid",
+    "dataset": {
+        "directory": "./",
+        "minimum_pub_date": 1483228800000,
+        "top_recent_pubs": 5
+    }
+}
+
```
mohituniyal, line 139: Missed an end quote here -> ```

README.md
```diff
@@ -103,11 +103,55 @@ Below you will find examples of possible configurations depending on the Model t
...
+    "exclusion_inv": "ICLR.cc/2020/Conference/-/Expertise_Selection",
+    "bid_inv": "ICLR.cc/2020/Conference/-/Add_Bid",
+    "dataset": {
+        "directory": "./",
+        "or": {
+            "minimum_pub_date": 1483228800000,
+            "top_recent_pubs": "10%"
+        }
+    }
+}
```
mohituniyal, line 154: End quote after this -> ```

README.md
```diff
@@ -78,7 +78,7 @@ There are two steps to detect duplicates:
 The dataset can be generated using the [OpenReview python API](https://github.com/openreview/openreview-py) which should be installed when this repository is installed. You can generate your own dataset from some other source as long as it is compliant with the format shown in the Datasets section.
 Start by creating an "experiment directory" (`experiment_dir`), and a JSON config file (e.g. `config.json`) in it. Go to the Configuration File section for details on how to create the `config.json`.
 
-Duplicate detection uses ELMo exclusively, since we normalize the scores for BM25. ELMo scores have values from 0 to 1. The closer a score is to 1, the more similar the submissions are.
+Duplicate detection uses ELMo exclusively, since we always normalize the scores for BM25. ELMo scores have values from 0 to 1. The closer a score is to 1, the more similar the submissions are. The `normalize` option for ELMo is disabled for duplicate detection.
```
mohituniyal, line 81: Thanks for the readme changes @carlosmondra . Can you also add an example with the `normalize` option. Right now I can see in the tests that this option is available but the readme does not mention how or where to use it.

carlosmondra: Sure!

carlosmondra: I have added an example and an explanation. Let me know what you think.

PR #43, opened 2020-05-05
Feature/emails instead of groupids

### LINKED ISSUE #45: Only count publications depending on the settings of the config
Currently, the title or the abstract can be used for computing affinity scores. However, when creating the dataset, even if the config says to only use the abstract, if the publication has no abstract in OpenReview the publication is still taken into account. We should only count the publications that match what was specified in the config file.

carlosmondra:
This PR adds the option to provide reviewer_ids as emails or as tilde_ids instead of the OpenReview groups. This way people outside OpenReview can extract the expertise of users.

Fixes #45


melisabok:
there are some conflicts here.

PR #46, opened 2020-06-05
execute the api calls in parallel

melisabok:
this is what I'm doing to get the expertise for NeurIPS reviewers.


melisabok:
this should be refactored a little bit, the function has too many parameters that can be reduced.

PR #52, opened 2020-10-09
Feature/create bid dataset

melisabok:


expertise/create_dataset.py
```diff
@@ -36,12 +36,31 @@ def convert_to_list(self, config_invitations):
         assert isinstance(invitations, list), 'Input should be a str or a list'
         return invitations
 
+    def get_paper_notes(self, author_id, dataset_params):
+
+        use_bids_as_expertise = dataset_params.get('bid_as_expertise', False)
+
+        if use_bids_as_expertise:
+            bid_invitation = dataset_params['bid_invitation']
+            bids = openreview.tools.iterget_edges(self.openreview_client, invitation=bid_invitation, tail=author_id)
```
carlosmondra, line 45: Maybe we should check if this value exists in the `dataset_params`. Otherwise, it will fail

carlosmondra, line 45: Is the `author_id` always going to be in the `tail`?

melisabok: I would like this to fail. If there is not bid invitation then no bid dataset can be created.

melisabok: Usually for bids, head is the paper id and tail is the profile id.

expertise/create_dataset.py
```diff
@@ -36,12 +36,31 @@ def convert_to_list(self, config_invitations):
...
+            note_ids = [e.head for e in bids if e.label in ['Very High', 'High']]
+
+            notes = self.openreview_client.get_notes_by_ids(ids=note_ids)
+            difference = list(set(note_ids) - set([n.id for n in notes]))
+            if difference:
+                print('difference', difference)
```
carlosmondra, line 51: Is this difference to check if there are papers that were desk rejected after the reviewer bid on them?

melisabok: Ohhh you are right, I should check the invitation id of each note!

expertise/create_dataset.py
```diff
@@ -36,12 +36,27 @@ def convert_to_list(self, config_invitations):
...
+    def get_paper_notes(self, author_id, dataset_params):
+
+        use_bids_as_expertise = dataset_params.get('bid_as_expertise', False)
+
+        if use_bids_as_expertise:
+            bid_invitation = dataset_params['bid_invitation']
+            paper_invitation = self.config['paper_invitation']
+            bids = openreview.tools.iterget_edges(self.openreview_client, invitation=bid_invitation, tail=author_id)
+            note_ids = [e.head for e in bids if e.label in ['Very High', 'High']]
+            return [n for n in self.openreview_client.get_notes_by_ids(ids=note_ids) if n.invitation == paper_invitation]
```
carlosmondra, line 48: Why can't it be like this?
```py
return [n for n in self.openreview_client.get_notes_by_ids(ids=note_ids)]
```

melisabok: I need to check if the note are not desk rejected or withdrawn.

carlosmondra (approved):

PR #68, opened 2021-06-01
Add paperhash for all papers to set

### LINKED ISSUE #66: Create dataset: ignore submissions/publications with empty title
We found out that a submission can be submitted with an empty title, just spaces. https://github.com/openreview/openreview-py/issues/913

Trim the title and check if it is empty then ignore it when creating the dataset.

### LINKED ISSUE #67: Create dataset: ignore publication that have a non null original value
Publications that were submitted to OpenReview in a double blind venue are listed twice in the profile publication section. We should take in account only one of them, the solution that implemented the Expertise Selection interface is ignoring the publications with an original value, only taking in account the original ones.

carlosmondra:
Always add to seen keys even if not added to the filtered papers

fixes #67
fixes #66

### DIFF
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -229,7 +229,8 @@ class OpenReviewExpertise(object):
 
             if n['id'] not in self.excluded_ids_by_user[member] and paperhash not in seen_keys:
                 filtered_papers.append(n)
-                seen_keys.add(paperhash)
+
+            seen_keys.add(paperhash)
 
         return member, email, filtered_papers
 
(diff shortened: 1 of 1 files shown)


expertise/create_dataset.py
```diff
@@ -227,9 +227,10 @@ def retrieve_expertise_helper(self, member, email):
 
             paperhash = openreview.tools.get_paperhash('', n['content']['title'])
 
-            if n['id'] not in self.excluded_ids_by_user[member] and paperhash not in seen_keys:
+            if n.get('original') is not None and n['id'] not in self.excluded_ids_by_user[member] and paperhash not in seen_keys:
```
melisabok, line 230: shouldn't be 'is None'? filtered_papers are the ones used for the expertise?

carlosmondra: you are right! Oh boy...

melisabok (approved):

PR #70, opened 2021-08-03
Run Expertise Model as Python Server

### LINKED ISSUE #69: Run expertise model as a service
We would like to provide a public API to compute affinity scores between papers and users, users and users or papers and papers. 

The API can be used by any user with an active OpenReview profile and the expertise scores will be computed with the data that is accesible to the user. 

Continue...

haroldrubio:
Resolves #69
This pull request adds the ability to host an API using a Python Flask server. Some notable changes so far include:

1. Moves code previously in `run.py` to a separate file `execute_expertise.py`. This was done to make the expertise model run with a single function call given a config - which is called both from the API and server, and also to ensure changes to expertise model code would affect both the command line and server versions of the code
2. Introduces several general purpose queue-like objects and two concrete implementations of these queues to interact with OpenReview structures and data
3. Adds tests `test_unit_queue.py` for the functionality of the general queues and `test_queue.py` for integrating the expertise model with the queues
4. Documents how to start the Flask server and a high-level overview of the created queue objects

Some things that still/may need to be worked on are:
- [ ] Redis integration for a more robust datastore of job metadata
- [ ] Store errors thrown by job processes that allow clients to understand why a job has failed during processing after submitting it
- [ ] Enabling some features of the queue system on the server API (e.g canceling jobs waiting in queue and enabling timeouts)
- [ ] Fix tests that are failing on CircleCI but not locally (most likely letting the tests sleep longer after submitting jobs to the queue)


expertise/execute_expertise.py
```diff
@@ -0,0 +1,177 @@
+from pathlib import Path
+import openreview
+from .create_dataset import OpenReviewExpertise
+from .dataset import ArchivesDataset, SubmissionsDataset, BidsDataset
+from .config import ModelConfig
+
+# Move run.py functionality to a function that accepts a config dict
+def execute_expertise(config_file=None, model_config=None):
```
melisabok, line 8: what is the difference between config_file and model_config?

expertise/execute_expertise.py
```diff
@@ -0,0 +1,177 @@
...
+            mfr_submissions_path=None,
+            scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '.csv')
+        )
+
+        if config['model_params'].get('sparse_value'):
+            ens_predictor.sparse_scores(
+                scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '_sparse.csv')
+            )
+
+def execute_create_dataset(client, config_file=None, model_config=None):
```
melisabok, line 166: is this functions create the dataset or running the model? I'm confused here.

expertise/execute_expertise.py
haroldrubio: The difference is `model_config` is already a ModelConfig object whereas `config_file` would be a dictionary that still needs to be turned into a ModelConfig object.
The different use cases are seen here, where the command line expects a JSON file, and uses the `config_file_path` argument:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/run.py#L10-L13
whereas the server directly reads a JSON/dictionary from the POST request:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/routes.py#L37
which later gets turned into a ModelConfig using the `config_dict` argument:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/execute_expertise.py#L12-L13

expertise/execute_expertise.py
```diff
@@ -0,0 +1,177 @@
...
+from .dataset import ArchivesDataset, SubmissionsDataset, BidsDataset
+from .config import ModelConfig
+
+# Move run.py functionality to a function that accepts a config dict
+def execute_expertise(config_file=None, model_config=None):
+
+    if model_config:
+        config = model_config
+    elif config_file:
+        config = ModelConfig(config_dict=config_file)
```
carlosmondra, line 13: What if you always expect `config_file` and create the ModelConfig object here instead of using if else statements?

expertise/execute_expertise.py
haroldrubio: In this file, the two functions `execute_expertise` corresponds to running the model, and `execute_create_dataset` corresponds to fetching data from OpenReview.

So everything from the following lines and below are moved to `execute_expertise`:
https://github.com/openreview/openreview-expertise/blob/b973858ce6434e3bbcfc473cda132dedaea36858/expertise/run.py#L12-L14

And everything from the following line and below is moved to `execute_create_dataset`:
https://github.com/openreview/openreview-expertise/blob/b973858ce6434e3bbcfc473cda132dedaea36858/expertise/create_dataset.py#L361

Which is used to define the work done by the queues:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/or_queue.py#L80-L82
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/or_queue.py#L158-L164

expertise/execute_expertise.py
haroldrubio: Ah, that does sound like a better solution - I'll make that change

expertise/service/README.md
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+
+---
+
+**`GET /results`**
+
+**Request Body:**
+
+`{ job_id: int, delete_on_get: boolean }`
```
carlosmondra, line 61: Can anyone retrieve the results of any job or is there a restriction?

expertise/service/README.md
```diff
@@ -0,0 +1,145 @@
...
+5.  Multi-threaded multi-process handling of submitted jobs
+6.  Allocating a directory on a per-job basis to enforce job isolation
+
+JobQueue defines several status codes that are free to asynchronously query during a job's lifetime:
+- `queued` -- The job is currently awaiting processing by a worker
+- `processing` -- The job is currently being worked on by a worker
+- `completed` -- The job has finished and the results are currently stored on the server
+- `stale` -- The job has been canceled before it arrived at processing
+- `timeout` -- The job has exceeded the specified timeout (there is no timeout by default)
+- `error` -- While processing, the job process has encountered an error
```
carlosmondra, line 111: This is a very impressive queue! I'm wondering if this should be in the definition of the class JobQueue instead of the README.

expertise/service/README.md
haroldrubio: At the moment, you can only access information about jobs that you have personally submitted:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/queue.py#L351-L353
Where the `id` is extracted from the token:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/routes.py#L42-L44

expertise/execute_expertise.py
melisabok: Got it! thanks for the explanation.

expertise/execute_expertise.py
melisabok: What if the only parameter is ModelConfig and it is created before calling execute_expertise?

expertise/execute_expertise.py
melisabok: I would define `execute_expertise( model_config=None)` where model_config can be created from a JSON that comes in the request body or from file.

expertise/execute_expertise.py
```diff
@@ -0,0 +1,177 @@
...
+    else:
+        raise Exception('Must provide either a model config or a dictionary of config parameters')
+    
+    archives_dataset = ArchivesDataset(archives_path=Path(config['dataset']['directory']).joinpath('archives'))
+    if Path(config['dataset']['directory']).joinpath('submissions').exists():
```
melisabok, line 18: I don't think the request.body should define this type of parameter. We have to internally decide which is the directory to store the csv files. 

I think the request body should define a subset of parameters supported in config.json

expertise/service/README.md
melisabok: Is the profile_id the id of the job?

melisabok: or do you show all the jobs associated to that user?

melisabok: I think we should have a way to query a single job id, like GET /jobs?jobId=[]

expertise/execute_expertise.py
haroldrubio: Yeah, this is something that I've been thinking about. What's currently implemented now is - the config gets pre-processed before being executed so that all directories are nested under it's own private directory. So for example if `config['dataset']['directory'] = '/test'`, this would be changed to `config['dataset']['directory'] = 'job_id/test'`:

https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/service/or_queue.py#L89-L95

But since it doesn't matter to the user what kind of structure the server's filesystem has, I was thinking of just overriding this completely.

I want to avoid modifying the code in the `execute_expertise` function, as the way I have changed the repo, this code is shared with the command line execution:
https://github.com/openreview/openreview-expertise/blob/a86bddea5e48d9efd20d00f8fe1a0f1f1be7e0f8/expertise/run.py#L13-L14

expertise/service/README.md
melisabok: I'm also impressed about the implementation!!!

I see the TwoStepQueue implementation is a little bit complex and difficult to follow. To make the things easier, I wouldn't worry about the token expired error and implement one single queue with no inner queues. 

This queue will have one single job that runs two tasks:

- execute_create_dataset
- execute_expertise

If then token is expired then we set a status error and the user should try again with a refreshed token.

Another option to avoid this error is to support username/password in the request headers and the API can instantiate the `openreview.Client` with a username and password instead of a token. 

Have you ever checked the Celery library to work with queues? https://docs.celeryproject.org/en/stable/getting-started/first-steps-with-celery.html#redis

I think this library can save you a lot of work. 

This is another option: https://python-rq.org/ but @purujitgoyal said that he had some issues with this one. 

I found another one: https://github.com/coleifer/huey

expertise/execute_expertise.py
melisabok: I think what you can do is to define a default ModelConfig object where you set all the directories and parameters and in the Flask controller you override some of them from the request body. I can say the parameters that can be sent in the body are:

- name (this must be a mandatory field)
- match_group (this must be a mandatory field)
- paper_invitation (this must be a mandatory field)
- exclusion_inv (optiona, default: None)
- model (optiona, default: SPECTER+MFR)
- model_params.use_title (optional, default: true)
- model_params.use_abstract  (optional, default: true)
- model_params.sparse_value (optional, default: None(?))
- model_params.average_score (optional, default: false)
- model_params.max_score (optional, default: true)
- model_params.skip_specter (optional: default: false)

@carlosmondra please take a look

So I think we don't need to follow the same format of config.json to define the request body parameters because the config model will be based on some internal options and user options so they must be translated anyway.

carlosmondra: There are some params that are model specific, so I'm not sure if we should throw an exception or just ignore them when parameters are passed that do not correspond to the model that is being used.
We cannot have max_score and average_score at the same time.

name (this must be a mandatory field) <--- I'm not sure if this is even necessary anymore. It depends what we want to use it for. A default name can be used if the param.
model (optional, default: SPECTER+MFR)
model_params.use_title (optional, default: true) <--- I'm not sure where this option is used other than BM25
model_params.use_abstract (optional, default: true) <--- I'm not sure where this option is used other than BM25
model_params.sparse_value (optional, default: None(?))
model_params.average_score (optional, default: false)
model_params.max_score (optional, default: true)
model_params.skip_specter (optional: default: false) <--- Not sure what this means. Maybe it's the separation we talked about between computing scores and computing embeddings

haroldrubio:
@melisabok 
Some things that have been giving me some trouble:

1. While testing the API, I need to manually copy the data in `tests/data/archives` to the job directory - this is because the `archives` folder in the job directory does not get properly populated with the mock client (this step occurs fine when testing with the live API on a data dump - this may be a side effect from me not fully understanding the data that is created in Carlos' mock).
In order to copy this data, I need to know the job directory ahead of time before POSTing an expertise request, so I cannot use the randomized ID that's implemented for a live API call.

2. I decided to keep the profile IDs in the working directories as a way of organizing which jobs were submitted by which profiles. If I just made directories based only on the job IDs, I would need some structure that keeps track of this information explicitly and that would also be tolerant to crashes - I'm not sure if this would be preferred over adding the profile IDs to the directories.

3. There was a suggestion to just pass the OpenReview client through the functions but I ran into an issue where the OpenReview client cannot be pickled (which is something done by Celery itself that I cannot change) so I believe we still need to log-in a new client when creating the dataset. But in an effort to keep the code cleaner, I've offloaded the mock client code to a `utils.py` file.

expertise/service/config/celery_config.py
```diff
@@ -0,0 +1,22 @@
...
+
+task_default_queue = 'default'
+task_queues = (
+    Queue('userpaper', routing_key='expertise.service.celery_tasks.run_userpaper'),
+    Queue('expertise', routing_key='expertise.service.celery_tasks.run_expertise')
+)
+# CELERY_IMPORTS = ('tasks')
+task_ignore_result = False
+# broker_url = 'redis://localhost:6379/0'
+broker_url = 'redis://localhost:6379/0'
```
melisabok, line 12: can you use a database number 10? it is the same the matching system is using.

expertise/service/config/default.cfg
```diff
@@ -0,0 +1,6 @@
+LOG_FILE='default.log'
+OPENREVIEW_BASEURL='http://localhost:3000'
+WORKING_DIR = 'TEMP_DIR'
```
melisabok, line 3: what is `TEMP_DIR`? a real folder name? I would use a default './tmp'

expertise/service/or_queue.py
```diff
@@ -0,0 +1,164 @@
+import os, openreview, shutil
```
melisabok, line 1: do you still need this file?

haroldrubio: Oh right I meant to remove the old queues - I can do that right now

tests/test_queue.py
```diff
@@ -0,0 +1,221 @@
...
+    test_num = server_config['TEST_NUM']
+    
+    if os.path.isdir(f'~{test_num}'):
+        shutil.rmtree(f'~{test_num}')
+
+    # Gather config
+    config = {
+        'name': 'test_run',
+        'match_group': ["ABC.cc"],
+        'dataset': {
```
melisabok, line 102: dataset shouldn't be passed as a parameter.

tests/test_queue.py
```diff
@@ -0,0 +1,221 @@
...
+        "model_params": {
+            "use_title": False,
+            "use_abstract": True,
+            "use_cuda": False,
+            "batch_size": 1,
+            "average_score": True,
+            "max_score": False,
+            "normalize": False,
+            "skip_elmo": False,
+            "scores_path": "./",
```
melisabok, line 116: all these path parameters are not needed here.

tests/test_queue.py
```diff
@@ -0,0 +1,221 @@
...
+        'match_group': ["ABC.cc"],
+        'dataset': {
+            'directory': 'tests/data/'
+        },
+        'csv_submissions': 'csv_submissions.csv',
+        "model": "elmo",
+        "model_params": {
+            "use_title": False,
+            "use_abstract": True,
+            "use_cuda": False,
```
melisabok, line 110: use_cuda also is not decided by the user, it can be a config of the server.

expertise/service/config/celery_config.py
haroldrubio: Sure that can be fixed

expertise/service/celery_tasks.py
```diff
@@ -0,0 +1,45 @@
...
+    try:
+        if 'token' in config.keys():
+            openreview_client = openreview.Client(
+                token=config['token'],
+                baseurl=config['baseurl']
+            )
+        else:
+            openreview_client = mock_client()
+            logger.info('Creating dataset')
+        execute_create_dataset(openreview_client, config_file=config)
```
melisabok, line 17: could you change the parameter name to `config`?

tests/test_queue.py
```diff
@@ -0,0 +1,221 @@
...
+        shutil.rmtree(f'~{test_num}')
+
+    # Gather config
+    config = {
+        'name': 'test_run',
+        'match_group': ["ABC.cc"],
+        'dataset': {
+            'directory': 'tests/data/'
+        },
+        'csv_submissions': 'csv_submissions.csv',
```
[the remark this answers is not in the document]
haroldrubio: `csv_submissions.csv` is used because there does not exist an invitation in the mock client that has corresponding submissions - so I copy over the data from the test directory into the job directory. 
Currently it's implemented as a feature of the API in test mode - the key `csv_submissions` gets overwritten to contain a list of CSV strings and those get written to the server.

I could look into making this done exclusively by the tests, or looking into inserting submissions + a submission invitation into the mock client.

expertise/service/routes.py
```diff
@@ -0,0 +1,390 @@
...
+    new_config['model_params']['specter_dir'] = flask.current_app.config['SPECTER_DIR']
+    new_config['model_params']['mfr_feature_vocab_file'] = flask.current_app.config['MFR_VOCAB_DIR']
+    new_config['model_params']['mfr_checkpoint_dir'] = flask.current_app.config['MFR_CHECKPOINT_DIR']
+    return new_config
+
+def enqueue_expertise(json_request, profile_id, in_test_mode):
+    # Throw error if required field isn't present
+    req_fields = ['name', 'match_group', 'paper_invitation']
+    for field in req_fields:
+        assert field in flask.request.json, f'Missing required field: {field}'
```
melisabok, line 85: it seems you are not using the parameter json_request here.

expertise/service/routes.py
```diff
@@ -0,0 +1,390 @@
...
+
+    if in_test_mode:
+        job_id = str(global_id.value)
+        global_id.value += 1
+    else:
+        job_id = ''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits + string.ascii_lowercase) for _ in range(5))
```
melisabok, line 91: what about using this to generate a job id https://pypi.org/project/shortuuid/?

expertise/service/celery_tasks.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            openreview_client = mock_client()
+            logger.info('Creating dataset')
+        execute_create_dataset(openreview_client, config=config)
+        run_expertise.apply_async(
+                (config, logger),
+                queue='expertise',
+        )
+    except Exception as exc:
+        working_dir = os.path.join(config['profile_dir'], str(config['job_id']))
```
melisabok, line 23: What is this for? is it possible to reduce the code in the exception block?

expertise/service/celery_tasks.py
```diff
@@ -0,0 +1,45 @@
...
+        logger.error('Error: {}'.format(exc))
+
+@celery.task(name='expertise', track_started=True, bind=True, time_limit=3600 * 24)
+def run_expertise(self, config: dict, logger: logging.Logger):
+    try:
+        logger.info('Executing expertise')
+        execute_expertise(config=config)
+        logger.info('Returned from expertise')
+    except Exception as exc:
+        working_dir = os.path.join(config['profile_dir'], str(config['job_id']))
```
melisabok, line 38: it seems here it is doing the same.

expertise/service/routes.py
```diff
@@ -0,0 +1,458 @@
...
+                token=token,
+                baseurl=flask.current_app.config['OPENREVIEW_BASEURL']
+            )
+            user_config['token'] = token
+            user_config['baseurl'] = flask.current_app.config['OPENREVIEW_BASEURL']
+        else:
+            openreview_client = mock_client()
+
+        profile, _ = get_profile_and_id(openreview_client)
+        job_id = enqueue_expertise(user_config, profile.id)
```
melisabok, line 227: is it possible to get the profile inside of the task? I would prefer to avoid the if test_mode here and only have it in a single place.

expertise/service/routes.py
```diff
@@ -0,0 +1,458 @@
...
+    except openreview.OpenReviewException as error_handle:
+        flask.current_app.logger.error(str(error_handle))
+
+        error_type = str(error_handle)
+        status = 500
+
+        if 'not found' in error_type.lower():
+            status = 404
+        elif 'forbidden' in error_type.lower():
+            status = 403
```
carlosmondra, line 241: The OpenReview exceptions don't have the status value?

expertise/service/routes.py
```diff
@@ -0,0 +1,458 @@
...
+        return flask.jsonify(result), status
+
+    # pylint:disable=broad-except
+    except Exception as error_handle:
+        result['error'] = 'Internal server error: {}'.format(error_handle)
+        return flask.jsonify(result), 500
+        
+    else:
+        flask.current_app.logger.debug('POST returns ' + str(result))
+        return flask.jsonify(result), 200
```
carlosmondra, line 361: This is a minor thing but why is this outside the `try:` block?

expertise/service/routes.py
```diff
@@ -0,0 +1,458 @@
...
+            result['results'] = ret_list
+
+            # Gather metadata
+            with open(metadata_dir, 'r') as metadata:
+                result['metadata'] = json.load(metadata)
+
+        # Clear directory
+        if delete_on_get:
+            flask.current_app.logger.error(f'Deleting {search_dir}')
+            shutil.rmtree(search_dir)
```
carlosmondra, line 435: Has it been discussed for how long we will store results if `delete_on_get` is not passed?

carlosmondra:
This is looking good! I think it's almost there :D

expertise/service/routes.py
haroldrubio: I don't think we've discussed this - maybe one solution is to keep them for a week?

Implementation-wise I was thinking I would store the creation date of the job and have a separate thread query all existing job directories for their creation date and if the current date is one week past, clean the directory. Then the thread would sleep for 24 hours so that this check would happen daily.

Is this a reasonable solution?

expertise/service/routes.py
haroldrubio: This gets executed only if the try block passes with no errors - this was just the way it was implemented in the original matcher code I pulled from.

Looking at it again, I could just put this code in the try block

expertise/service/routes.py
haroldrubio: Status here refers to the HTTP status code to be returned, not necessarily trying to find the status value of the OpenReview exception:
https://github.com/openreview/openreview-expertise/blob/fa8c9ce32653858df385f2d1e66baae9e71418da/expertise/service/routes.py#L244

expertise/service/celery_tasks.py
haroldrubio: These exception blocks are made to log any error made when running either `execute_create_dataset` or `execute_expertise`.

I can trim down the size of these blocks by removing some extra logger statements.

expertise/service/expertise.py
```diff
@@ -99,6 +99,8 @@ def _prepare_config(self, request):
             os.makedirs(config['dataset']['directory'])
         with open(os.path.join(root_dir, 'config.json'), 'w+') as f:
             ## Remove the token before saving this in the file system
+            if 'token' in config.keys():
+                del config['token']
```
melisabok, line 105: are you sure that you don't need the token later?

haroldrubio:
I made two requests mainly to see if the server could handle more than one job - though I don't do thorough testing of the outcome of the second job (as in I don't check if it ever completes). I could either add in some code to test the outcome of the second job, or we could just test 1 successful request for now.

expertise/service/expertise.py
```diff
@@ -0,0 +1,303 @@
...
+
+        :returns file_dir: The directory of the score file, if it exists, starting from the given directory
+        :returns metadata_dir: The directory of the metadata file, if it exists, starting from the given directory
+        """
+        # Search for scores files (only non-sparse scores)
+        file_dir, metadata_dir = None, None
+        with open(os.path.join(search_dir, 'config.json'), 'r') as f:
+            config = json.load(f)
+        # Look for score files
+        for root, dirs, files in os.walk(search_dir, topdown=False):
```
purujitgoyal, line 166: Why are you searching the whole directory? You can directly look for file using `os.path.exists`

expertise/service/expertise.py
```diff
@@ -0,0 +1,303 @@
...
+        :returns: A list of subdirectories not prefixed by the given root directory
+        """
+        subdirs = [name for name in os.listdir(self.working_dir) if os.path.isdir(os.path.join(self.working_dir, name))]
+        if user_id is None:
+            return subdirs
+        else:
+            # If given a profile ID, assume looking for job dirs that contain a config with the
+            # matching profile id
+            filtered_dirs = []
+            for subdir in subdirs:
```
purujitgoyal, line 143: is this scalable?

expertise/service/expertise.py
```diff
@@ -0,0 +1,303 @@
...
+                config_dir = os.path.join(self.working_dir, subdir, 'config.json')
+                with open(config_dir, 'r') as f:
+                    config = json.load(f)
+                    if user_id == config['user_id'] or user_id.lower() in SUPERUSER_IDS:
```
purujitgoyal, line 147: You can add a special case for superuser to return all subdirectories directly.

expertise/service/expertise.py
```diff
@@ -0,0 +1,303 @@
...
+                    })
+            result['results'] = ret_list
+
+            # Gather metadata
+            with open(metadata_dir, 'r') as metadata:
+                result['metadata'] = json.load(metadata)
+
+        # Clear directory
+        if delete_on_get:
+            self.logger.error(f'Deleting {search_dir}')
```
purujitgoyal, line 300: `self.logger.info`

tests/test_expertise_service.py
```diff
@@ -0,0 +1,451 @@
...
+        # response = test_client.get('/expertise/status', query_string={'id': f'{job_id}'}).json['results']
+        # assert len(response) == 1
+        # assert response[0]['name'] == 'test_run'
+        # assert response[0]['status'] == 'Queued'
+        # assert response[0]['description'] == 'Server received config and allocated space'
+
+        # Query until job is complete
+        response = test_client.get('/expertise/status', query_string={'id': f'{job_id}'}).json['results']
+        assert len(response) == 1
+        while response[0]['status'] != 'Completed':
```
purujitgoyal, line 211: This can be an infinite loop. Fix the number of tries and mark the test as failure after that.

tests/test_expertise_service.py
```diff
@@ -0,0 +1,451 @@
...
+        assert openreview_context['job_id'] is not None
+        test_client = openreview_context['test_client']
+        # Query until job is err
+        time.sleep(5)
+        response = test_client.get('/expertise/results', query_string={'id': f"{openreview_context['job_id']}"})
+        assert response.status_code == 404
+
+        response = test_client.get('/expertise/status', query_string={'id': f"{openreview_context['job_id']}"}).json['results']
+        assert len(response) == 1
+        while response[0]['status'] == 'Processing':
```
purujitgoyal, line 288: This can be an infinite loop. Fix the number of tries and mark the test as failure after that.

tests/test_expertise_service.py
```diff
@@ -0,0 +1,451 @@
...
+        assert openreview_context['job_id'] is not None
+        id_list = openreview_context['job_id']
+        num_requests = len(id_list)
+        test_client = openreview_context['test_client']
+        last_job_id = id_list[num_requests - 1]
+
+        # Assert that the last request completes
+        response = test_client.get('/expertise/status', query_string={'id': f'{last_job_id}'}).json['results']
+        assert len(response) == 1
+        while response[0]['status'] != 'Completed':
```
purujitgoyal, line 351: This can be an infinite loop. Fix the number of tries and mark the test as failure after that.

expertise/service/expertise.py
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+        # Define expected/required API fields
+        req_fields = ['name', 'match_group', 'paper_invitation', 'user_id', 'job_id']
+        optional_model_params = ['use_title', 'use_abstract', 'average_score', 'max_score', 'skip_specter']
+        optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl']
+        path_fields = ['work_dir', 'scores_path', 'publications_path', 'submissions_path']
+
+        # Validate + populate fields
+        for field in req_fields:
```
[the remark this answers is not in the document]
haroldrubio: Ahh I see - that's a change that I can make

expertise/execute_expertise.py
```diff
@@ -0,0 +1,168 @@
...
+        if config['model_params'].get('sparse_value'):
+            ens_predictor.sparse_scores(
+                scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '_sparse.csv')
+            )
+
+def execute_create_dataset(client, config=None):
+
+    config = ModelConfig(config_dict=config)
+    
+    print(config)
```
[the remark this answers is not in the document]
haroldrubio: There is no logger object here as this function is also called from the command line program. The logger at the moment only exists in the server-level abstraction

expertise/service/expertise.py
```diff
@@ -77,24 +79,47 @@ def _prepare_config(self, request):
         optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl']
         path_fields = ['work_dir', 'scores_path', 'publications_path', 'submissions_path']
 
-        # Validate + populate fields
+        # Populate fields
```
purujitgoyal, line 82: This is great. One last thing, take out the validation steps as a function.

expertise/service/expertise.py
purujitgoyal: @haroldrubio If we start using Redis to store the jobIDs and their major properties, we won't need to save directories for the jobs, right?

expertise/service/celery_tasks.py
```diff
@@ -0,0 +1,62 @@
+from functools import update_wrapper
+import logging, json, os, shutil
+from .utils import mock_client
+from expertise.execute_expertise import *
```
purujitgoyal, line 4: Don't use `*` to import

expertise/service/expertise.py
```diff
@@ -0,0 +1,423 @@
...
+                "use_title": True,
+                "batch_size": 4,
+                "use_abstract": True,
+                "average_score": False,
+                "max_score": True,
+                "skip_specter": False,
+                "use_cuda": False
+            }
+        }
+        # Define expected/required API fields
```
purujitgoyal, line 76: These can move to `init`

expertise/service/expertise.py
```diff
@@ -0,0 +1,434 @@
...
+        self.working_dir = config['WORKING_DIR']
+        self.specter_dir = config['SPECTER_DIR']
+        self.mfr_feature_vocab_file = config['MFR_VOCAB_DIR']
+        self.mfr_checkpoint_dir = config['MFR_CHECKPOINT_DIR']
+
+        # Define expected/required API fields
+        self.req_fields = ['name', 'match_group', 'paper_invitation', 'user_id', 'job_id']
+        self.optional_model_params = ['use_title', 'use_abstract', 'average_score', 'max_score', 'skip_specter']
+        self.optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl']
+        self.path_fields = ['work_dir', 'scores_path', 'publications_path', 'submissions_path']
```
purujitgoyal, line 67: `path_fields` is not used anywhere, do we need this? If not, remove it.

PR #77, opened 2021-11-08
Fix API functionality

### LINKED ISSUE #73: Expertise API: Move default expertise config to server config
Currently, the default configuration run by the expertise API resides in a function in `expertise.py`. This makes it difficult to access key model parameters that may need to vary as the machine/demands of the API changes. This should be moved to Flask config file for ease-of-access.

### LINKED ISSUE #74: Expertise API: Standardize error/status messages
Currently, there are several different formats for errors across different endpoints: some exceptions have the format `{'error': 'some_error'}` or `{'results:[{'status':'Error', 'description': 'some_error'}]}`

Change the format of error and status messages to have the form `{'status': 'some_status', 'description': 'some_description'}`

### LINKED ISSUE #81: Expertise API: Cannot delete error jobs
Usually, the way to delete a job is to retrieve the results from `expertise/results` and set the `delete_on_get` flag so that the following lines are run:
https://github.com/openreview/openreview-expertise/blob/4e53d71b5fc33d61e36b488bfc788c4ef8faa198/expertise/service/expertise.py#L446-L450

However, because an exception is raised before those lines are run, then the block to delete the job is never executed:
https://github.com/openreview/openreview-expertise/blob/4e53d71b5fc33d61e36b488bfc788c4ef8faa198/expertise/service/expertise.py#L424-L426

Add a way to delete a job that hasn't completed yet. This can either be done by either 1) moving this block above the exception to allow for deletion before checking for scores or 2) introducing a new endpoint `expertise/delete` dedicated to removing the job from the server.

### LINKED ISSUE #82: Expertise API: Two jobs of the same name are ambiguous
Two jobs of the same name can be submitted since only the `job_id` is unique. This can make distinguishing between which job was submitted when, difficult. Add a new field `mdate` to track when the status of a job has been updated. Return this new `mdate` and the existing `cdate` to the `expertise/status` endpoint.

haroldrubio:
Closes #73. Closes #74. Closes #82. Closes #81.

This solves the following leftover issues with the expertise API which are:
- Moving the default job config to the Flask config file
- Standardizes the error formatting to match the message returned from the `/expertise/status` endpoint

### DIFF
--- a/expertise/service/README.md
+++ b/expertise/service/README.md
@@ -22,9 +22,26 @@ Returns code `200` on successful job submission with: `{"job_id": "string"}`\
 Returns code `400` if there was an error in the submitted config
 
 ## `GET /expertise/status`
-This endpoint gets the status of all jobs submitted by the user, or a single job with the given `job_id`. A valid request body has the following format:
+This endpoint gets the status of a single job with the given `job_id`. A valid request body has the following format:
 ```
-{"id": "string" [Optional]}
+{"id": "string"}
+```
+
+Returns the status, if any, that were submitted by the user and has the job id:
+```
+{
+	"job_id": "string",
+	"name": "string",
+	"status": "string",
+	"description": "string",
+	"config": {...}
+}
+```
+
+## `GET /expertise/status/all`
+This endpoint gets the status of all jobs submitted by the user. A valid request body has the following format:
+```
+{}
 ```
 
 Returns a list of jobs, if any, that were submitted by the user and has the job id if provided:
@@ -35,7 +52,8 @@ Returns a list of jobs, if any, that were submitted by the user and has the job 
         "job_id": "string",
         "name": "string",
         "status": "string",
-        "description": "string"
+        "description": "string",
+        "config": {...}
       }
   ]
 }
--- a/expertise/service/celery_tasks.py
+++ b/expertise/service/celery_tasks.py
@@ -25,7 +25,7 @@ def update_status(job_dir, new_status, desc=None):
         config['description'] = descriptions[new_status]
     else:
         config['description'] = desc
-    config['mdate'] = int(time.time())
+    config['mdate'] = int(time.time() * 1000)
     with open(os.path.join(config['job_dir'], 'config.json'), 'w+') as f:
         json.dump(config, f, ensure_ascii=False, indent=4)
 
--- a/expertise/service/celery_tasks.py
+++ b/expertise/service/celery_tasks.py
@@ -1,5 +1,5 @@
 from functools import update_wrapper
-import logging, json, os, shutil
+import logging, json, os, shutil, time
 from .utils import mock_client
 from expertise.execute_expertise import execute_create_dataset, execute_expertise
 from expertise.service.expertise import JobStatus, JobDescription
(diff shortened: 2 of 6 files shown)


haroldrubio:
That's a good idea! I hadn't thought of that, sounds like I'll have to play around with breaking the openreview-api haha

melisabok:
@carlosmondra could you share an example of an error JSON the openreview-api returns so Harold can use the same schema?

carlosmondra:
> @carlosmondra could you share an example of an error JSON the openreview-api returns so Harold can use the same schema?

```js
{
  name: 'NotFoundError' // All the errors end with Error
  message: 'The requested page was not found' // This is the message that is shown to the user
  status: 404
  details: { path: '/nonexistentpage', user: '~Carlos_Mondragon1' } // The details object can contain anything that is relevant to the error. It does not need to contain path and user necessarily; this is just an example :).
}
```

melisabok (approved):

purujitgoyal (dismissed):

expertise/service/README.md
```diff
@@ -22,9 +22,26 @@ Returns code `200` on successful job submission with: `{"job_id": "string"}`\
...
+This endpoint gets the status of a single job with the given `job_id`. A valid request body has the following format:
 ```
-{"id": "string" [Optional]}
+{"id": "string"}
+```
+
+Returns the status, if any, that were submitted by the user and has the job id:
+```
+{
+	"job_id": "string",
```
carlosmondra, line 33: JSON fields should be camel-cased and should only contain underscores at the beginning of the field. You can read more about it [here](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format)

expertise/service/celery_tasks.py
```diff
@@ -25,6 +25,7 @@ def update_status(job_dir, new_status, desc=None):
         config['description'] = descriptions[new_status]
     else:
         config['description'] = desc
+    config['mdate'] = int(time.time())
```
carlosmondra, line 28: We save the timestamps in milliseconds. I think this is saving it in seconds though

expertise/service/config/default.cfg
```diff
@@ -5,3 +5,16 @@ SPECTER_DIR = '../expertise-utils/specter/'
...
+    "model": "specter+mfr",
+    "model_params": {
+        "use_title": True,
+        "batch_size": 4,
+        "use_abstract": True,
+        "average_score": False,
+        "max_score": True,
+        "skip_specter": False,
+        "use_cuda": False
+    }
```
carlosmondra, line 19: I'm realizing that in the POST request you send these parameters as well and they have underscores. Well, I guess we can leave the change for a different PR.

expertise/service/routes.py
```diff
@@ -29,6 +29,37 @@ def get_client():
...
+        error_name = 'ForbiddenError'
+    elif status_code == 404:
+        error_name = 'NotFoundError'
+    elif status_code == 500:
+        error_name = 'InternalServerError'
+
+    template = {
+        'name': error_name,
+        'message': description,
+        'status': status_code
```
carlosmondra, line 58: Is having `status` necessary?

haroldrubio: I don't believe it is - I think it is redundant to have it here again

expertise/service/celery_tasks.py
haroldrubio: Ah I see an example of this is the openreview-py repo, this is a simple fix

carlosmondra (approved):

PR #78, opened 2021-11-10
Affinity Scoring for Journal Venues

### LINKED ISSUE #75: Run expertise model for a single paper against a group of users
For journal venues, individual papers are reviewed on a rolling basis instead of large batches under a single invitation. Add a new function for the expertise model to run affinity scoring for one paper vs a group. Also, create an endpoint in the expertise API for this function.

haroldrubio:
Closes #75.

Adds a new field `paper_id` to the expertise model's configuration file which allows the user to fetch a single paper by ID, and to score that paper against a group of reviewers. Additionally, this functionality is also enabled for the expertise API.

- Adds a new field `paper_id` to the expertise model's configuration file
- Enable `paper_id` feature for the expertise API as an optional field for the `/expertise` endpoint
- Changes the `paper_invitation` field for `/expertise` endpoint from required to optional
- Patches a bug for `specter+mfr` that made it crash on a single submission by duplicating the submission under a dummy ID

Branched off of #77 to add the new returned statuses, so that must be merged first.


expertise/create_dataset.py
```diff
@@ -289,12 +290,23 @@ def retrieve_expertise(self):
...
             submissions.extend(list(openreview.tools.iterget_notes(
                 self.openreview_client, invitation=invitation_id)))
 
+        if paper_id:
+            submissions.append(self.openreview_client.get_note(paper_id))
+
+        # Bug: specter+mfr cannot handle a single submission
+        # Solution: create a copy of the note and modify the ID
+        if self.config.get('model') == 'specter+mfr' and len(submissions) == 1:
+            dummy = deepcopy(submissions[0])
```
melisabok, line 306: instead of adding a library for copying one single submissions, could you just create the new submission with the properties that we already know the submission has? id, title, abstract.

melisabok:
I'm realizing the TMLR Journal uses the API v2 that has a different domain and the note schema is slightly different. I suggest two options:

1. Support in the body the request a submission object, it should contain the id, content.title, content.abstract. 
2. Support a new parameter `version:2` that requests two different domains and should read the data differently. 

Option 1 is easier to implement but Option 2 must be implemented at some point.

melisabok (approved):

expertise/create_dataset.py
```diff
@@ -289,12 +290,23 @@ def retrieve_expertise(self):
...
         for invitation_id in invitation_ids:
             submissions.extend(list(openreview.tools.iterget_notes(
                 self.openreview_client, invitation=invitation_id)))
 
+        if paper_id:
+            submissions.append(self.openreview_client.get_note(paper_id))
+
+        # Bug: specter+mfr cannot handle a single submission
+        # Solution: create a copy of the note and modify the ID
+        if self.config.get('model') == 'specter+mfr' and len(submissions) == 1:
```
purujitgoyal, line 305: What's the exact issue here @haroldrubio? Can we modify specter or mfr code to handle single submission?

expertise/service/expertise.py
```diff
@@ -43,9 +43,9 @@ def __init__(self, client, config, logger):
         self.mfr_checkpoint_dir = config['MFR_CHECKPOINT_DIR']
 
         # Define expected/required API fields
-        self.req_fields = ['name', 'match_group', 'paper_invitation', 'user_id', 'job_id']
+        self.req_fields = ['name', 'match_group', 'user_id', 'job_id']
         self.optional_model_params = ['use_title', 'use_abstract', 'average_score', 'max_score', 'skip_specter']
-        self.optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl']
+        self.optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl', 'paper_invitation', 'paper_id']
```
purujitgoyal, line 48: I'm assuming one of either `paper_invitation` or `paper_id` is required, right @haroldrubio? If yes, add a check for that in `validate_config`.

tests/test_expertise_service.py
```diff
@@ -87,15 +87,15 @@ def test_request_expertise_with_no_config(self, openreview_context, celery_sessi
         assert response.status_code == 400, f'{response.json}'
         assert 'Error' in response.json['name']
         assert 'bad request' in response.json['message'].lower()
-        assert response.json['message'] == 'Bad request: missing required field: name match_group paper_invitation'
+        assert response.json['message'] == 'Bad request: missing required field: name match_group'
```
purujitgoyal, line 90: Add a test when both `paper_invitation` and `paper_id` are missing

expertise/create_dataset.py
haroldrubio: There was an error in the model code where something was being calculated incorrectly, specifically this assertion was failing:
https://github.com/openreview/openreview-expertise/blob/4e53d71b5fc33d61e36b488bfc788c4ef8faa198/expertise/models/multifacet_recommender/multifacet_recommender.py#L903

While you were away, we discussed it with Haw Shiuan and he agreed that we could just duplicate the submission as a simple fix.

purujitgoyal (approved):

PR #83, opened 2021-11-22
Affinity Scoring between Groups

### LINKED ISSUE #76: Run the expertise model for one group of users against another group
For some cases, we will want to compute scores for one group against another (i.e area chairs vs senior area chairs). In order to do so, fetch the publications for 1 group and place them in `submissions.json`. For the other group, place their publications in an `archives` folder. Lastly, perform some averaging across the raw scores from the expertise model. Also, add this functionality under a new expertise API endpoint.

haroldrubio:
Closes #76.

This PR adds a new String field to the configuration file called `submission_group`.

- The publications of the members of `submission_group` are stored in `submissions.json`
- The mapping between publications and the members is stored in `publications_by_profile_id.json`.
- The `submission_group` field triggers a post-processing step where submission scores are averaged

Branched off of #77 to add the new returned statuses, so that must be merged first.


tests/test_expertise_service.py
```diff
@@ -528,6 +528,81 @@ def test_fetch_high_load_results(self, openreview_context, celery_session_app, c
...
+        # Submit a working job and return the job ID
+        MAX_TIMEOUT = 600 # Timeout after 10 minutes
+        test_client = openreview_context['test_client']
+        # Make a request
+        response = test_client.post(
+            '/expertise',
+            data = json.dumps({
+                    'name': 'test_run',
+                    'match_group': ["ABC.cc"],
+                    'submission_group': 'ABC.cc',
```
melisabok, line 541: I'm not sure about `submission_group` name, we are trying to support the computation of scores of groups vs group. We should hide to the user the implementation about using one of the group as submissions. 

Why the `match_group` is an array? What happens if there is more than one element?

Can we use match_group array to represent both groups? first element is the match_group and the second element is the other group represented as submissions.

what do you think?

expertise/create_dataset.py
```diff
@@ -300,12 +300,50 @@ def retrieve_expertise(self):
...
+        client = self.openreview_client
+        group_a_members = client.get_group(submission_group).members
+        pbar = tqdm(total=len(group_a_members), desc=submission_group + ' status...')
+        publications_by_profile_id = {}
+        all_papers = []
+
+        def get_status(profile_id):
+            pbar.update(1)
+            profile = openreview.tools.get_profile(client, profile_id)
+            if profile:
```
purujitgoyal, line 313: Handle the case if there's no profile

expertise/execute_expertise.py
```diff
@@ -158,6 +158,56 @@ def execute_expertise(config):
                 scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '_sparse.csv')
             )
 
+    if 'submission_group' in config:
```
purujitgoyal, line 161: I'd suggest moving this whole part into a function and using it inside the `all_scores` function for each model type.

expertise/service/expertise.py
```diff
@@ -464,23 +463,34 @@ def get_expertise_results(self, user_id, job_id, delete_on_get=False):
...
+                            continue
+
+                        ret_list.append({
+                            'submission': row[0],
+                            'user': row[1],
+                            'score': float(row[2])
+                        })
+                result['results'] = ret_list
+            else:
+                # If submission group, group under different keys
```
purujitgoyal, line 484: why different keys?

expertise/execute_expertise.py
```diff
@@ -158,6 +158,56 @@ def execute_expertise(config):
...
+        for author_file in archive_files:
+            archive_members.append(author_file.split('.')[0])
+
+        # Fetch submission members
+        with open(dataset_root.joinpath('publications_by_profile_id.json'), 'r') as f:
+            publications_by_profile_id = json.load(f)
+        submission_members = publications_by_profile_id.keys()
+
+        # Perform averaging
+        average_score = {}
```
purujitgoyal, line 188: Not really sure, but this averaging part can be made more efficient using the pandas `grouping_by` method. Take a look and see if it helps or not.

tests/test_expertise_service.py
haroldrubio: Sure, I can rename it from `submission_group` to something more fitting - we can discuss this.

`match_group` has always been able to accept a list of group IDs - in the documentation for the config file:
`"match_group": ["ICLR.cc/2020/Conference/Reviewers", "ICLR.cc/2020/Conference/Area_Chairs"],`
which just retrieves all the profiles from all group IDs to fetch their publications. I could change this to use your suggestion but I would have to move the functionality of grabbing from multiple groups somewhere else.

expertise/create_dataset.py
haroldrubio: What do you suggest happens when there is no profile? Throw an exception? Admittedly, this part of the code is something Carlos wrote for ICLR so I'm not too familiar with these sections of code.

expertise/service/expertise.py
haroldrubio: What I mean by this comment is not creating tuples of `(submission, user, score)`, create tuples of `(match_member, submission_member, score)` so these keys have to be different to be explicit on what scores are being returned.

expertise/execute_expertise.py
haroldrubio: What if sparse scores are being computed between groups? Then this step will be missed - do you think I should also add it into the `sparse_scores` functions as well?

Also, this block of code requires access to the `dataset` field of the config, which is not passed to the models so I would have to add an additional argument to these functions. Though I do agree that this block should probably be in a function.

expertise/create_dataset.py
melisabok: if there is no profile then ignore it, there won't be scores for that user.

expertise/create_dataset.py
```diff
@@ -300,12 +305,55 @@ def retrieve_expertise(self):
...
+            result = future.result()
+            # Convert publications to json
+            papers_json_list = []
+            for paper_note in result['papers']:
+                papers_json_list.append(paper_note.to_json())
+            publications_by_profile_id[result['profile_id']] = papers_json_list
+            all_papers = all_papers + result['papers']
+
+        # Dump publications by profile id
+        with open(self.root.joinpath('publications_by_profile_id.json'), 'w') as f:
```
melisabok, line 337: is this file being stored in the job folder or somewhere else?

shouldn't be called submission.json?

expertise/create_dataset.py
```diff
@@ -300,12 +305,55 @@ def retrieve_expertise(self):
...
+        return all_papers
 
     def get_submissions(self):
         invitation_ids = self.convert_to_list(self.config.get('paper_invitation', []))
         paper_id = self.config.get('paper_id')
+        submission_group = None
         submissions = []
 
+        # If there is more than 1 group, set the second group as the submission group
+        group_ids = self.convert_to_list(self.config.get('match_group', []))
```
melisabok, line 349: isn't `match_group` being used when we compute scores between reviewers and submissions? 

is this going to load all the papers for all the reviewers?

expertise/create_dataset.py
```diff
@@ -256,6 +256,11 @@ def retrieve_expertise(self):
         use_email_ids = self.config.get('use_email_ids', False)
         group_ids = self.convert_to_list(self.config.get('match_group', []))
         reviewer_ids = self.convert_to_list(self.config.get('reviewer_ids', []))
+
+        # If there is more than 1 group, only take the first group
+        if len(group_ids) > 1:
```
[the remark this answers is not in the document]
haroldrubio: Yes, the `match_group` field can either be a string or an array. In this function, the string gets converted into an array with a single element in the `self.convert_to_list()` function.

Here, I only take the first group because the first group corresponds to members whose publications will be dumped into the `archives` folder. The second group corresponds to members whose publications will be dumped into `submissions.json`

expertise/create_dataset.py
haroldrubio: Ah so this is a mapping that is maintained only for group scoring. The reason this is created is because the expertise model expects a list of submissions with no author metadata in the `submissions.json` file. So, we have to keep track of which author wrote which paper in the `submissions.json` file to do the averaging post-processing step. This is stored in a file because the expertise model is called separately after `create_dataset`, so it cannot just remain in memory.

expertise/create_dataset.py
haroldrubio: A reviewer-submission scoring job would have fields `match_group: ['ABC.cc'], paper_invitation: 'ABC.cc/-/Submission'` whereas a group-group scoring job would have no paper_invitation and only have `['ABC.cc/Senior_Area_Chairs', 'ABC.cc/Area_Chairs']`

So when there are at least 2 groups present in the `match_group` array, the way this is adapted into the reviewer-submission framework is that the first group acts as the reviewers and all publications of all members in the second group acts as the submissions

README.md
```diff
@@ -180,7 +180,7 @@ Below you will find examples of possible configurations depending on the Model t
 
 ### Create Dataset Configuration Options
 This parameters could be included in a separate file, like `dataset-config.json`, as was mentioned before.
-- `match_group`: (optional) String or array of strings containing the groups of Reviewers or Area Chairs. The Reviewers (and Area Chairs) will get affinity scores with respect to the submitted papers based on their expertise. This expertise is obtained based on the publications available in OpenReview. It can be provided instead of or on top of the `reviewer_ids` and `csv_expertise`, the values will be combined.
+- `match_group`: (optional) String or array of strings containing groups IDs. The group will get affinity scores with respect to the submitted papers based on their expertise. This expertise is obtained based on the publications available in OpenReview. If 2 group IDs are given, the members of the two groups will be matched based on their expertise. It can be provided instead of or on top of the `reviewer_ids` and `csv_expertise`, the values will be combined.
```
melisabok, line 183: could you update this?

expertise/create_dataset.py
```diff
@@ -300,12 +301,53 @@ def retrieve_expertise(self):
...
+        # Dump publications by profile id
+        with open(self.root.joinpath('publications_by_profile_id.json'), 'w') as f:
+            json.dump(publications_by_profile_id, f, indent=2)
+        
+        return all_papers
 
     def get_submissions(self):
         invitation_ids = self.convert_to_list(self.config.get('paper_invitation', []))
         paper_id = self.config.get('paper_id')
+        submission_groups = self.convert_to_list(self.config.get('alternate_match_group', []))
```
melisabok, line 344: we need to be very careful here where either paper_invitation or alternate_match_group are posible but not both at the same time.

melisabok (approved):

expertise/create_dataset.py
purujitgoyal: When there's no profile, `get_status` will return `None`, which will cause problems later on when you try to access results from this `None` object. (line 330)

expertise/create_dataset.py
haroldrubio: Sorry I missed this comment! Do you suggest I validated this in the `get_submissions` function or should I just ensure that the API does not submit a config with both of these fields?

melisabok: I would prefer to validate this in the API request class.  But in the get_submissions implementation don't load the submissions from paper_invitation and alternate_match_group. Use one only:

if invitation_ids:
    ## load the papers from the invitations
else:
   ## load the papers from alternate_match_group

melisabok (approved):

expertise/create_dataset.py
purujitgoyal: @haroldrubio did you fix this?

haroldrubio: Yes, this function gets called here:
https://github.com/openreview/openreview-expertise/blob/454706ba16972a8d9bbaee8e066a9f1531fc5dd0/expertise/create_dataset.py#L324-L326

so `group_a_members` gets set here:
https://github.com/openreview/openreview-expertise/blob/454706ba16972a8d9bbaee8e066a9f1531fc5dd0/expertise/create_dataset.py#L308-L311

which uses `get_profile_ids()` defined here which guarantees that the members returned in the first element of the tuple have profiles:
https://github.com/openreview/openreview-expertise/blob/454706ba16972a8d9bbaee8e066a9f1531fc5dd0/expertise/create_dataset.py#L149

purujitgoyal (approved):

PR #85, opened 2021-12-10
Support Camel Case Fields in API

haroldrubio:
This PR allows fields previously separated by underscores to be written using camel case to adhere to JSON styling conventions.


expertise/service/expertise.py
```diff
@@ -48,6 +48,17 @@ def __init__(self, client, config, logger):
         self.optional_fields = ['model', 'model_params', 'exclusion_inv', 'token', 'baseurl', 'baseurl_v2', 'paper_invitation', 'paper_id']
         self.path_fields = ['work_dir', 'scores_path', 'publications_path', 'submissions_path']
 
+        # For the optional fields, add camel case equivalents
+        camel_case_params = []
```
melisabok, line 52: instead of converting the field names, why don't you use the camel version directly?

melisabok: I think we can create a Config class that can be serialized from JSON object, similar to openreview-py classes.(Note, Invitation, etc)

purujitgoyal: +1

purujitgoyal: > I think we can create a Config class that can be serialized from JSON object, similar to openreview-py classes.(Note, Invitation, etc)

+1

haroldrubio:
For the validation of the fields in the request, should I return a bad request when extra/unexpected information is included even if it not used?
When reading from the entity objects, any extra information is never read and stored in the final config. However, for the model object, all extra information gets store in the final config - I just decided to allow this since the expertise system only reads from specific fields anyway, but maybe this should be changed?

This validation also differs from the current validation because it can only return 1 error in the request at a time whereas previously it returned all errors with the request. Should I look into keeping the feature where all errors are returned at the same time?

expertise/service/utils.py
```diff
@@ -158,3 +159,166 @@ def get_user_id(openreview_client):
...
+                self.paper_invitation = inv
+            elif 'id' in entity.keys():
+                id = _get_from_entity('id')
+                self.paper_id = id
+            else:
+                raise openreview.OpenReviewException(f"Bad request: no valid {type} properties in {entity_id}")
+        else:
+            raise openreview.OpenReviewException(f"Bad request: invalid type in {entity_id}")
+
+    def from_request(self, request):
```
melisabok, line 233: why don't you call it `from_json`?

haroldrubio: I called it `from_request` because the dict that it expects has the new API format (camel case fields), where as the `to_json` function converts the class into a dict that follows a format that is accepted by the expertise model (underscore/snake fields). I made this distinction to allow for a `from_json` function that is able to accept a dict formatted for the expertise model.

However, I'm open to changing it to `from_json` since I think this class will only be used for validation and loading from the initial API call.

expertise/service/expertise.py
```diff
@@ -79,11 +80,13 @@ def _prepare_config(self, request) -> dict:
...
-        config = self._validate_fields(request)
-        self.logger.info(f"Config validation passed - setting server-side fields")
+        validated_request = APIRequest()
+        validated_request.from_json(request)
+        config = ServerConfig(self._get_default_config(), validated_request).to_json()
+        self.logger.info(f"Config validation passed - setting server-side fields in {config}")
 
         # Populate with server-side fields
-        root_dir = os.path.join(self.working_dir, request['job_id'])
+        root_dir = os.path.join(self.working_dir, config['job_id'])
```
melisabok, line 89: if config is an instance of `ServerConfig` , shouldn't be `config.job_id` here?

expertise/service/expertise.py
```diff
@@ -102,98 +105,33 @@ def _prepare_config(self, request) -> dict:
             config['model_params']['mfr_checkpoint_dir'] = self.mfr_checkpoint_dir
 
         # Create directory and config file
+        token = config.pop('token')
```
melisabok, line 108: what does it mean `pop` here?

expertise/service/expertise.py
```diff
@@ -102,98 +105,33 @@ def _prepare_config(self, request) -> dict:
...
-            if len(error_fields['required']) > 0:
-                error_string += 'missing required field: ' + ' '.join(error_fields['required']) + '\n'
-            if len(error_fields['unexpected']) > 0:
-                error_string += 'unexpected field: ' + ' '.join(error_fields['unexpected']) + '\n'
-            if len(error_fields['model_params']) > 0:
-                error_string += 'unexpected model param: ' + ' '.join(error_fields['model_params']) + '\n'
-            raise OpenReviewException(error_string.strip())
-        else:
-            return config
+        return config, token
```
melisabok, line 114: it seems config already has a token, why do you return it here?

expertise/service/utils.py
```diff
@@ -158,3 +159,234 @@ def get_user_id(openreview_client):
...
+        def _camel_to_snake(camel_str):
+            camel_str = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', camel_str)
+            return re.sub('([a-z0-9])([A-Z])', r'\1_\2', camel_str).lower()
+        
+        # Get the name of the job
+        self.name = _get_field_from_request('name')
+
+        # These are automatically provided by the server
+        self.user_id = _get_field_from_request('user_id')
+        self.job_id = _get_field_from_request('job_id')
```
melisabok, line 212: why the request has a job_id?

expertise/service/utils.py
```diff
@@ -158,3 +159,234 @@ def get_user_id(openreview_client):
...
+                raise openreview.OpenReviewException(f"Bad request: no valid {type} properties in {entity_id}")
+        else:
+            raise openreview.OpenReviewException(f"Bad request: invalid type in {entity_id}")
+
+        # Check for extra entity fields
+        if len(source_entity.keys()) > 0:
+            raise openreview.OpenReviewException(f"Bad request: unexpected fields in {entity_id}: {list(source_entity.keys())}")
+
+
+class ServerConfig(object):
```
melisabok, line 266: I'm realizing that this configuration is attached to the job that is running, why don't we call it `JobConfig` instead?

expertise/service/expertise.py
haroldrubio: `config` gets the JSON value of `ServerConfig` because the expertise model only accepts a dict. I could for example, change the model so that it can load from the fields of a `ServerConfig` (`JobConfig` from your later comment) and that would let us use a `JobConfig` object throughout for consistency.

expertise/service/expertise.py
haroldrubio: `pop` gets the value of the token and removes the key-value pair from the dict

To also address your Line 114 comment, this is so the token never gets written to disk as part of the config, and the token gets passed as an argument to the Celery task

expertise/service/utils.py
haroldrubio: I agree that would be a more fitting name for this class

expertise/service/utils.py
haroldrubio: I can probably improve this. Some of these metadata fields get set `routes.py`, some get set in `start_expertise()` and `_prepare_config()` in `expertise.py` but I should move all of these fields into a single spot

expertise/service/expertise.py
melisabok: Now, I understand why you can do this, you are calling to_json after creating the JobConfig object, why?

Can't you just use the config object? then when you need to run the model you can save the config object to a json file.

melisabok: you can also move this code to the JobConfig class:

```
        # Set SPECTER+MFR paths
        if 'specter' in config.get('model', 'specter+mfr'):
            config['model_params']['specter_dir'] = self.specter_dir
        if 'mfr' in config.get('model', 'specter+mfr'):
            config['model_params']['mfr_feature_vocab_file'] = self.mfr_feature_vocab_file
            config['model_params']['mfr_checkpoint_dir'] = self.mfr_checkpoint_dir

```

expertise/service/expertise.py
```diff
@@ -79,11 +80,19 @@ def _prepare_config(self, request) -> dict:
                         when it is not expected
         """
         # Validate fields
-        config = self._validate_fields(request)
-        self.logger.info(f"Config validation passed - setting server-side fields")
+        validated_request = APIRequest()
+        validated_request.from_json(request)
+        config = JobConfig(self._get_default_config(), validated_request).to_json()
```
melisabok, line 85: default_config can be loaded inside of the class, no need to passot as a parameter

haroldrubio: Hmm I would like to do this but the JobConfig class has no knowledge of the Flask configuration so it would not know where to look to load this data

haroldrubio:
Most of the meaningful changes in `expertise.py` happen above and in `start_expertise()` - the code below that function is mostly migrating to using `JobConfig` objects.

I've also reformatted the `JobConfig` class a bit to mimic similar classes in openreview-py.

expertise/service/expertise.py
```diff
@@ -208,128 +108,41 @@ def _get_score_and_metadata_dir(self, search_dir):
...
         run_userpaper.apply_async(
-            (config, self.logger),
+            (config.to_json(), token, self.logger),
             queue='userpaper',
             task_id=job_id
         )
-        with open(os.path.join(config['job_dir'], 'config.json'), 'w+') as f:
-            json.dump(config, f, ensure_ascii=False, indent=4)
+        self.logger.info(f"\nconf: {config.to_json()}\n")
+        with open(os.path.join(config.job_dir, 'config.json'), 'w+') as f:
```
melisabok, line 144: can you add a method `save_config` in JobConfig class? it seems you are saving the file in different places.

expertise/service/expertise.py
```diff
@@ -79,121 +57,43 @@ def _prepare_config(self, request) -> dict:
...
-        If the server has detected any errors in the proposed config, assemble the error and raise the exception
-        Otherwise, this function returns the properly augmented config
-
-        :param request: Contains the initial request from the user
-        :type request: dict
-
-        :returns config: A server-compatible version of the user request
+        if not os.path.isdir(config.dataset['directory']):
+            os.makedirs(config.dataset['directory'])
+        with open(os.path.join(config.job_dir, 'config.json'), 'w+') as f:
```
melisabok, line 73: same here, you can to `config.save()`

melisabok:
I like the new changes! I made one last comment, @purujitgoyal could you do a final review?

melisabok (dismissed):

expertise/service/README.md
```diff
@@ -30,10 +45,10 @@ This endpoint gets the status of a single job with the given `job_id`. A valid r
 Returns the status, if any, that were submitted by the user and has the job id:
```
purujitgoyal, line 45: line 42 would be `{"id": string}`, same as above.

expertise/service/expertise.py
```diff
@@ -208,128 +107,40 @@ def _get_score_and_metadata_dir(self, search_dir):
...
         with open(os.path.join(search_dir, 'config.json'), 'r') as f:
-            config = json.load(f)
+            config = JobConfig.from_json(json.load(f))
 
         # Look for files
-        if os.path.isfile(os.path.join(search_dir, f"{config['name']}.csv")):
-            file_dir = os.path.join(search_dir, f"{config['name']}.csv")
+        if os.path.isfile(os.path.join(search_dir, f"{config.name}.csv")):
+            file_dir = os.path.join(search_dir, f"{config.name}.csv")
         if file_dir is None:
```
purujitgoyal, line 115: Could be changed to single `if-else` rather than two `if`

expertise/service/expertise.py
```diff
@@ -208,128 +107,40 @@ def _get_score_and_metadata_dir(self, search_dir):
...
-            raise OpenReviewException("Score file not found for job {job_id}".format(job_id=config["job_id"]))
+            raise OpenReviewException("Score file not found for job {job_id}".format(job_id=config.job_id))
 
         if os.path.isfile(os.path.join(search_dir, 'metadata.json')):
             metadata_dir = os.path.join(search_dir, 'metadata.json')
         if metadata_dir is None:
```
purujitgoyal, line 120: Same as above, single `if-else`

expertise/service/expertise.py
```diff
@@ -357,21 +168,21 @@ def get_expertise_all_status(self, user_id):
             self.logger.info(f"Attempting to load {search_dir}/config.json")
             with open(os.path.join(search_dir, 'config.json'), 'r') as f:
                 s = f"{''.join(f.readlines())}"
-                config = json.loads(s)
-            status = config['status']
-            description = config['description']
+                config = JobConfig.from_json(json.loads(s))
```
purujitgoyal, line 171:
```suggestion
                config = JobConfig.from_json(json.load(f))
```

No need to convert to string first.

expertise/service/expertise.py
```diff
@@ -408,20 +219,20 @@ def get_expertise_status(self, user_id, job_id):
         self.logger.info(f"Attempting to load {search_dir}/config.json")
         with open(os.path.join(search_dir, 'config.json'), 'r') as f:
             s = f"{''.join(f.readlines())}"
-            config = json.loads(s)
-        status = config['status']
-        description = config['description']
+            config = JobConfig.from_json(json.loads(s))
```
purujitgoyal, line 222:
```suggestion
            config = JobConfig.from_json(json.load(f))
```

expertise/service/expertise.py
```diff
@@ -525,13 +335,12 @@ def del_expertise_job(self, user_id, job_id):
         self.logger.info(f"Attempting to load {search_dir}/config.json")
         with open(os.path.join(search_dir, 'config.json'), 'r') as f:
             s = f"{''.join(f.readlines())}"
-            config = json.loads(s)
+            config = JobConfig.from_json(json.loads(s))
```
purujitgoyal, line 338:
```suggestion
            config = JobConfig.from_json(json.load(f))
```

haroldrubio:
Looking at your feedback Purujit, I saw that there was a mismatch between sending an `id` in the request and returning a `job_id` in the response in other places as well.

I've standardized the other endpoints to require `job_id` in their requests, tested it with my openreview-py branch, and updated the documentation for those in the `expertise/service` README.

purujitgoyal (approved):

PR #88, opened 2021-12-14
Upgrade Expertise to API V2

haroldrubio:
This PR adds support for the OpenReview API V2 by using a different `openreview-py` client and adjusting for a different `Note` schema. Some of the following changes are:

- Adds a new keyword `version` to the `create_dataset` task
- Fetch data differently from `Note V2` based on `version`
- Replace the client used for the Celery queue task for V2
- Adds a new mock data file with the relevant fields updated to V2
- Adds tests for the upgraded `create_dataset` task and a journal-like test for the Expertise API


README.md
```diff
@@ -190,6 +190,7 @@ This parameters could be included in a separate file, like `dataset-config.json`
 - `bid_inv` (optional): String or array of strings with the bid invitations. Bids are used by the reviewers in OpenReview to select papers that they would or would not like to review. These bids are then used to compute a final affinity score to be more fair with the reviewers.
 - `use_email_ids` (optional): Boolean value. If true, then the email of the user is used instead of his/her OpenReview Profile ID.
 - `max_workers` (optional): Number indicating the amount of workers that will be used to retrieve the expertise from OpenReview. If it is not set, it will use the maximum available workers of your machine by default. The more workers the faster the creation of the dataset
+- `version` (optional): An integer indicating the version of the OpenReview API to use
```
purujitgoyal, line 193: mention here that default is set to 1

expertise/service/expertise.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, client, config, logger):
         self.mfr_checkpoint_dir = config['MFR_CHECKPOINT_DIR']
 
         # Define expected/required API fields
-        self.req_fields = ['name', 'match_group', 'user_id', 'job_id']
+        self.req_fields = ['name', 'match_group', 'user_id', 'job_id', 'version']
```
purujitgoyal, line 46: Don't make it a required field, let it be optional as mentioned in Readme.

purujitgoyal:
Add an exception for every `if-else` block if version is not 1 or 2.

README.md
melisabok: yes, please version 1 should be the default one

expertise/create_dataset.py
```diff
@@ -75,12 +81,21 @@ def get_publications(self, author_id):
                 continue
             if getattr(publication, 'cdate') is None:
                 publication.cdate = getattr(publication, 'tcdate', 0)
+
+            # Get title + abstract depending on API version
+            if self.get_api_version() == 1:
+                pub_title = publication.content.get('title')
+                pub_abstr = publication.content.get('abstract')
+            elif self.get_api_version() == 2:
+                pub_title = publication.content.get('title').get('value')
```
melisabok, line 90:
```suggestion
                pub_title = publication.content.get('title', {}).get('value')
```

expertise/create_dataset.py
```diff
@@ -75,12 +81,21 @@ def get_publications(self, author_id):
...
+                pub_abstr = publication.content.get('abstract').get('value')
```
melisabok, line 91:
```suggestion
                pub_abstr = publication.content.get('abstract', {}).get('value')
```

expertise/create_dataset.py
```diff
@@ -311,11 +326,20 @@ def get_submissions(self):
         reduced_submissions = {}
         for paper in tqdm(submissions, total=len(submissions)):
             paper_id = paper.id
+
+            # Get title + abstract depending on API version
+            if self.get_api_version() == 1:
+                paper_title = paper.content.get('title')
+                paper_abstr = paper.content.get('abstract')
+            elif self.get_api_version() == 2:
+                paper_title = paper.content.get('title').get('value')
```
melisabok, line 335: same here, use `{}` as a default value where the key is not found

expertise/create_dataset.py
```diff
@@ -48,7 +51,10 @@ def get_paper_notes(self, author_id, dataset_params):
             note_ids = [e.head for e in bids if e.label in ['Very High', 'High']]
             return [n for n in self.openreview_client.get_notes_by_ids(ids=note_ids) if n.invitation == paper_invitation]
 
-        return openreview.tools.iterget_notes(self.openreview_client, content={'authorids': author_id})
+        if self.get_api_version() == 1:
+            return openreview.tools.iterget_notes(self.openreview_client, content={'authorids': author_id})
+        elif self.get_api_version() == 2:
+            return openreview.tools.iterget_notes(self.openreview_client, content={'authorids': { 'value': [author_id] }})
```
melisabok, line 57: no need to make this change, the API 2 supports `content={'authorids': author_id}`. 

Check this call from the dev site:

`notes = list(openreview.tools.iterget_notes(dev_client, content={ 'authorids': '~Melisa_Bok1'}))`

melisabok:
I'm realizing that passing a version number won't work. There are notes that are going to be in the V1 and another ones in the V2 because they are saved in different collections. 

I think the expertise API should try to get the expertise and the submissions from both APIs and merge the results. 

to get the expertise:
```
notes_api1 = list(openreview.tools.iterget_notes(api1_client, content={ 'authorids': '~Melisa_Bok1'}))

notes _api2= list(openreview.tools.iterget_notes(api2_client, content={ 'authorids': '~Melisa_Bok1'}))
```
expertise = notes_api1 + notes_api2

to get the submissions:
```
submissions = list(openreview.tools.iterget_notes(api1_client, invitation=ICLR.cc/202/Confenrece/-/Blind_Submission))

if not submissions:
       submissions = list(openreview.tools.iterget_notes(api2_client, invitation=ICLR.cc/202/Confenrece/-/Blind_Submission))
```
the same happens for paper_id. All the note that belong to a venue should be saved in the same collection.

haroldrubio:
@melisabok @purujitgoyal I believe the PR is ready for another review. Something to note is that for testing purposes - if there is no token in the client, then I load a mock V2 client instead of logging in a real V2 client, similar to what is done in `celery_tasks.py`

expertise/create_dataset.py
```diff
@@ -294,11 +313,22 @@ def get_submissions(self):
...
         for invitation_id in invitation_ids:
-            submissions.extend(list(openreview.tools.iterget_notes(
-                self.openreview_client, invitation=invitation_id)))
+            submissions_v1 = list(openreview.tools.iterget_notes(
+                self.openreview_client, invitation=invitation_id))
+
+            if not submissions_v1:
+                submissions.extend(list(openreview.tools.iterget_notes(
+                    self.openreview_client_v2, invitation=invitation_id)))
+            else:
```
purujitgoyal, line 322: Don't need this `else`, just extend it before the `if`, even if it's empty, it won't matter.

expertise/create_dataset.py
```diff
@@ -294,11 +313,22 @@ def get_submissions(self):
...
+                submissions.extend(submissions_v1)
 
         if paper_id:
-            submissions.append(self.openreview_client.get_note(paper_id))
+            submission_v1 = self.openreview_client.get_note(paper_id)
+
+            if not submission_v1:
+                submissions.append(self.openreview_client_v2.get_note(paper_id))
+            else:
+                submissions.append(submission_v1)
```
purujitgoyal, line 331: Same as above.

purujitgoyal:
@haroldrubio Add a test for the wrong version. Everything else looks fine.

expertise/create_dataset.py
haroldrubio: Do you know what the behavior of the `openreview-py` client is when attempting to fetch notes that don't exist? I'm fine with this change if a failed `get_notes` and `get_note` call results in something like empty lists.

The current behavior of the mock client in the tests returns `None` in this situation, which breaks the tests because the following line fails:
https://github.com/openreview/openreview-expertise/blob/60b9a077f3bb16a03cc639e23d3aae261741d254/expertise/create_dataset.py#L313

But I can adjust the mock client to be more accurate to the live Python client if I know what happens when trying to get a note/notes that don't exist.

haroldrubio:
> @haroldrubio Add a test for the wrong version. Everything else looks fine.

Can you explain this further? The only place the `version` argument appears at the moment is when initializing the mock clients - `create_dataset` attempts to call both V1 and V2 so I'm not sure where this test would fit in.

expertise/create_dataset.py
purujitgoyal: No, but I assumed it'd return an empty list same as when you use `openreview_client_v2`. @melisabok can you confirm this?

melisabok: get_notes return a list
get_note return a Note object or an exception if the note doesn't exist

you can use tools.get_note that returns a None if the note doesn't exist.

when querying the notes by invitation we should assume the result will come from one of the APIs.

expertise/create_dataset.py
```diff
@@ -15,10 +15,27 @@
 from tqdm import tqdm
 from concurrent.futures import ThreadPoolExecutor
 from collections import defaultdict
+from expertise.service.utils import mock_client
 
 class OpenReviewExpertise(object):
-    def __init__(self, openreview_client, config):
+    def __init__(self, openreview_client, config, baseurl_v2=None):
```
melisabok, line 21: Why don't you pass the openreview_client_v2 as a parameter here?

expertise/create_dataset.py
```diff
@@ -71,16 +89,26 @@ def get_publications(self, author_id):
                 if not 'title' in publication.content or not publication.content.get('title'):
                     continue
             # Exclude blind Notes
-            if getattr(publication, 'original') is not None:
+            if hasattr(publication, 'original') and getattr(publication, 'original') is not None:
```
purujitgoyal, line 92: can be shortened to `getattr(publication, 'original', None)`

expertise/create_dataset.py
```diff
@@ -294,11 +322,21 @@ def get_submissions(self):
...
+                self.openreview_client, invitation=invitation_id))
+
+            submissions.extend(submissions_v1)
             submissions.extend(list(openreview.tools.iterget_notes(
-                self.openreview_client, invitation=invitation_id)))
+                self.openreview_client_v2, invitation=invitation_id)))
 
         if paper_id:
-            submissions.append(self.openreview_client.get_note(paper_id))
+            try:
```
purujitgoyal, line 335: Not a good way to use `try-except`. It could fail because of other exceptions as well. Use a separate `try-except` block for both API calls and only catch the exception for NoteNotFound.

purujitgoyal: Also, throw an exception if Note is not found by both API calls.

expertise/create_dataset.py
melisabok: I didn't know you could do that!

expertise/create_dataset.py
haroldrubio: I had originally done that but thought that too many things needed to be changed - but I can pass a V2 client instead since the V1 client is being passed in already

purujitgoyal (approved):

melisabok (approved):

purujitgoyal (approved):

PR #90, opened 2022-01-18
Add script to clean SPECTER temp files

haroldrubio:
This PR adds a simple script that clears the temp directory (as determined by Python) of sub-directories that are used by SPECTER. Specifically, when the `load_archive` function is called from the AllenNLP library, it creates a copy of the SPECTER model.

This script will not be necessary once AllenNLP gets updated, which automatically cleans up the temp files immediately.


README.md
```diff
@@ -153,6 +153,11 @@ MFR_CHECKPOINT_DIR = '../expertise-utils/multifacet_recommender/mfr_model_checkp
 
 See `/expertise/service/README.md` for documentation on the API design and endpoints.
 
+When SPECTER is run as a service, it creates ~1GB of extra files get stored in a temporary directory and are not automatically cleaned up. In order to cleanup these files, you can run the following command. **WARNING**: Any currently running jobs have the potential to fail when performing this operation
+```
+python -m expertise.clean_tmp
```
melisabok, line 158: Should we call it like this from the CRON job?

purujitgoyal:
Rather than using a python script, we can directly use a simple bash script, like this:

`find /tmp/* -prune -type d -mtime +1 -exec rm -r {} \;`

This will remove all the directories in `tmp` folder which were created more than a day ago.

melisabok:
@haroldrubio do we need this to be installed before TMLR?

clean_tmp.sh
```diff
@@ -0,0 +1 @@
+find /tmp/* -prune -type d -mtime +0.084 -exec rm -r {} \;
```
purujitgoyal, line 1: @haroldrubio do you know is there a particular pattern to directory names which get created? I'm worried removing everything from `/tmp` might cause problems elsewhere

haroldrubio: Yup the directories start with `tmp`, so `/tmp/tmp71atajqf`. Is would `/tmp/tmp*` be the correct regex?

purujitgoyal: I believe so.

purujitgoyal (approved):

melisabok (approved):

PR #99, opened 2022-03-08
Expertise API: Migrate configs to Redis

haroldrubio:
This PR changes what is stored in a job's configuration, moves its storage location from directly on the filesystem to Redis, and updates the return fields of some endpoints.


expertise/service/expertise.py
```diff
@@ -70,7 +75,10 @@ def _prepare_config(self, request) -> dict:
         # Create directory and config file
         if not os.path.isdir(config.dataset['directory']):
             os.makedirs(config.dataset['directory'])
-        config.save()
+        with open(os.path.join(config.job_dir, 'config.json'), 'w+') as f:
+            json.dump(config.to_json(), f, ensure_ascii=False, indent=4)
+        self.logger.info(f"Saving processed config to {os.path.join(config.job_dir, 'config.json')}")
+        config.save(self.redis_args)
```
purujitgoyal, line 81: why are you saving `config.json` in both memory and redis?

expertise/service/expertise.py
```diff
@@ -315,8 +288,9 @@ def get_expertise_results(self, user_id, job_id, delete_on_get=False):
 
         # Clear directory
         if delete_on_get:
-            self.logger.info(f'Deleting {search_dir}')
-            shutil.rmtree(search_dir)
+            self.logger.info(f'Deleting {config.job_dir}')
+            shutil.rmtree(config.job_dir)
+            JobConfig.remove_job(user_id, job_id, self.redis_args)
```
purujitgoyal, line 293: Rather than explicitly removing the job config from Redis, you can just set an expiry date for each config key.

purujitgoyal: Never mind, It's for `delete_on_get`, but either way, I think we should add an expiration date for config, something like a week or two? What do you think @melisabok?

purujitgoyal:
@haroldrubio Regarding creating and passing a Redis connection object everytime, you can use Redis connection pool. I just pushed a [commit](https://github.com/openreview/openreview-expertise/pull/108/commits/b8f1502c899eaed9b1b69d6e427c165e74a65fc7) to the embedding caching branch. I think that should work here as well. So you won't have to pass the `redis_args` everywhere.

expertise/service/expertise.py
melisabok: after the job is finished?

haroldrubio:
I just saw the commit, that is interesting - I'll have to play around with that. The connection pool lets me import the Redis connection args rather than having to pass them to the Celery task per request.

expertise/service/expertise.py
haroldrubio: I think we should still save some record of the job config on the disk and not just in Redis, for administrative purposes. The big difference is that this file is written once and never interacted with again by the API

expertise/service/expertise.py
haroldrubio: I think this is a good idea, but I am concerned about the mismatch between configs in Redis and its related files on the disk. Maybe Redis has a handler that is run when an object expires that also lets us clear the job data?

melisabok: Right, Redis has a property called `ttl` to specify the expiration date. 

But I'm not sure if I want to keep the job data in the files, aren't we going to clean the jobs folder regularly?

expertise/service/celery_tasks.py
```diff
@@ -23,43 +26,68 @@ def update_status(config, new_status, desc=None):
     else:
         config.description = desc
     config.mdate = int(time.time() * 1000)
-    config.save()
+    redis_db = RedisDatabase(connection_pool=redis_conn_pool)
+    redis_db.save_job(config)
+
+def on_failure_userpaper(self, exc, task_id, args, kwargs, einfo):
+    config, logger = args[0], args[2]
```
melisabok, line 33: could you import the logger here instead of passing it as an argunment?

expertise/service/expertise.py
```diff
@@ -177,7 +182,7 @@ def get_expertise_all_status(self, user_id, query_params):
                 self._filter_config(config)
                 result['results'].append(
                     {
-                        'job_id': job_dir,
+                        'job_id': config.job_dir,
```
melisabok, line 185: why don't you return `config` here? it seems you are making a new dictionary with the values from config plus config converted as a json, isn't that redudant?

tests/data/example_config.json
```diff
@@ -0,0 +1,54 @@
+{
+    "name": "test_run",
+    "user_id": "test_user1@mail.com",
+    "job_id": "BaxDa",
+    "baseurl": "http://localhost:3000",
```
melisabok, line 5: why do you need to store this here? I think this should be part of the API config not the job config.

tests/data/example_config.json
```diff
@@ -0,0 +1,54 @@
...
+    "baseurl_v2": "http://localhost:3000",
+    "job_dir": "./tests/jobs/BaxDa",
```
melisabok, line 7: same here `job_dir` shouldn't be data release to the user, it should be something that we do internally. the API can build this directory using the job_id plus a server configuration that specifies the jobs dir.

tests/data/example_config.json
```diff
@@ -0,0 +1,54 @@
...
+        "use_title": false,
+        "use_abstract": true,
+        "average_score": true,
+        "max_score": false,
+        "skip_specter": false,
+        "specter_batch_size": 16,
+        "mfr_batch_size": 50,
+        "use_cuda": false,
+        "name": "specter+mfr",
+        "work_dir": "./tests/jobs/BaxDa",
```
melisabok, line 28: same here. I would avoid all these directories here. They should be part of the server configuration.

tests/data/example_config.json
haroldrubio: I made this config file as mock data for one of the test, but I realized that I can just store an empty JobConfig object so I'll just remove this file.

But I will keep in mind your comments here and below in picking which fields gets stored on disk using the `from_json` function.

I have another PR that I'm working on that changes the status messages returned by the API to return the request rather than the job config.

melisabok: I see, what should we do? merge this and review the other PR?

haroldrubio: That other PR can wait until we merge both of the Redis PRs, I don't think it a priority

expertise/service/celery_tasks.py
haroldrubio: I tried directly to import the Flask current app logger but it seems like I would have to make additional changes elsewhere to make Celery aware of Flask's current app and I'm not sure if that would be worth it

melisabok:
Is this code already in master? Should we close it or resolve the conflicts?

purujitgoyal:
Closing. Merged with #108.

PR #103, opened 2022-03-21
Add status search to /all endpoint

haroldrubio:
Search all jobs by status by providing a prefix of the desired status

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -172,21 +172,19 @@ class ExpertiseService(object):
             status = config.status
             description = config.description
 
-            if search_status and not status.lower().startswith(search_status.lower()):
-                continue
-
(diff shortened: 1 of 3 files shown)


expertise/service/expertise.py
```diff
@@ -170,7 +171,10 @@ def get_expertise_all_status(self, user_id):
                 config = JobConfig.from_json(json.load(f))
             status = config.status
             description = config.description
-            
+
+            if search_status and not status.lower().startswith(search_status.lower()):
```
melisabok, line 175: Could you avoid the `continue` sentence?. 

```
if !search_status or status.lower().startswith(search_status.lower()):
    results['results'].append(...)
```

melisabok (approved):

PR #108, opened 2022-04-18
Embeddings Caching

purujitgoyal:


melisabok:
There are some tests failing here.

expertise/models/multifacet_recommender/specter.py
```diff
@@ -58,10 +61,13 @@ def __init__(self,
...
                  batch_size: int,
                  print_to_console: bool,
-                 has_dataset_reader: bool) -> None:
+                 has_dataset_reader: bool,
+                 store_redis: bool = False) -> None:
         super(_PredictManagerCustom, self).__init__(predictor, input_file, output_file, batch_size, print_to_console,
                                                     has_dataset_reader)
         self.total_size = int(sum([1 for _ in open(self._input_file)]) / self._batch_size)
+        self._store_redis = store_redis
+        self._redis_con = redisai.Client(host='localhost', port=6379, db=5)
```
melisabok, line 70: is it possible to pass the Redis client as a parameter in the constructor?

expertise/models/multifacet_recommender/specter.py
```diff
@@ -154,14 +177,15 @@ def set_archives_dataset(self, archives_dataset):
...
                 self.pub_note_id_to_abstract[publication['id']] = publication['content'].get('abstract', "")
-                if publication['id'] in output_dict:
-                    output_dict[publication['id']]["authors"].append(profile_id)
-                else:
-                    paper_ids_list.append(publication['id'])
-                    output_dict[publication['id']] = {"title": self.pub_note_id_to_title[publication['id']],
-                                                      "abstract": self.pub_note_id_to_abstract[publication['id']],
-                                                      "paper_id": publication["id"],
-                                                      "authors": [profile_id]}
+                if not self.redis.exists(publication['id']):
```
melisabok, line 180: what happens if the publication is in redis?

expertise/models/multifacet_recommender/specter.py
```diff
@@ -304,14 +329,34 @@ def load_emb_file(emb_file):
             print(len(bad_id_set))
             return emb_tensor, id_list, bad_id_set
 
-        with open(publications_path) as f_in:
-            print('Loading cached publications...')
-            paper_emb_train, train_id_list, train_bad_id_set = load_emb_file(f_in)
-            paper_num_train = len(train_id_list)
+        def load_from_redis():
+            paper_emb_size_default = 768
```
melisabok, line 333: what is `768` for?

expertise/models/multifacet_recommender/specter.py
```diff
@@ -304,14 +329,34 @@ def load_emb_file(emb_file):
...
+            id_list = self.pub_note_id_to_title.keys()
+            emb_list = []
+            bad_id_set = set()
+            con = redisai.Client(host='localhost', port=6379, db=5)
+            for paper_id in id_list:
+                try:
+                    paper_emb = con.tensorget(key=paper_id, as_numpy_mutable=True)
```
melisabok, line 340: why do you use tensorget to get the embeddings by a key that is a string?

expertise/models/multifacet_recommender/specter.py
```diff
@@ -304,14 +329,34 @@ def load_emb_file(emb_file):
...
+                    assert len(paper_emb) == paper_emb_size_default
+                    emb_list.append(paper_emb)
+                except Exception as e:
+                    bad_id_set.add(paper_id)
+
+            emb_tensor = torch.tensor(emb_list, device=torch.device('cpu'))
```
melisabok, line 346: why do you do this? it doesn't seem to be part of the previous code. Do you need this for caching?

expertise/utils/utils.py
```diff
@@ -510,6 +511,65 @@ def aggregate_by_group(config):
...
+            pub_dict = json.loads(pub_string)
+            archive_map[member_id].append(pub_dict.get('id', ''))
+    with open(f"{cache_dir}/archives.json", 'w') as f:
+        json.dump(archive_map, f)
+
+    # Copy work files - MFR is stored in the work_dir, SPECTER is stored in pub2vec
+    shutil.copytree(f"{work_dir}/mfr", f"{cache_dir}/work")
+    shutil.copy2(f"{pub_dir}/pub2vec.jsonl", f"{cache_dir}")
+    
+def from_cache(model_name, group_id, work_dir, pub_dir, dataset_dir):
```
melisabok, line 545: do you still need this? I thought that all the caching was going to happen through Redis. Why do you need to cache the files?

expertise/models/multifacet_recommender/specter.py
purujitgoyal: I believe so, but I'd like to keep this client here itself, as this is `RedisAI` client which will only be used by specter.

tests/test_expertise_service.py
```diff
@@ -472,7 +472,7 @@ def test_request_journal(self, openreview_context, celery_session_app, celery_se
                     },
                     "entityB": { 
                         'type': "Note",
-                        'id': 'KHnr1r7H'
+                        'id': '[KHnr1r7H]'
```
melisabok, line 475: I'm not sure why this is an array now.

expertise/models/multifacet_recommender/specter.py
purujitgoyal: Nothing, we don't need to calculate embedding for this paper in that case, and specter will get these embeddings when running the `all_scores` method.

expertise/models/multifacet_recommender/specter.py
purujitgoyal: It's the default embedding size that BERT returns.

expertise/models/multifacet_recommender/specter.py
purujitgoyal: It's there in the previous code as well. Check [`load_emb_file`](https://github.com/openreview/openreview-expertise/blob/a2a552a6e028e23fab81866bde29328fc46171db/expertise/models/multifacet_recommender/specter.py#L310)

expertise/utils/utils.py
purujitgoyal: They are not used anymore, will remove these.

expertise/models/multifacet_recommender/specter.py
melisabok: yes, I see that, but where do you decide how to load the embeddings? calling to either `load_from_redis` or `load_emb_file`?

purujitgoyal: [Here](https://github.com/openreview/openreview-expertise/blob/a2a552a6e028e23fab81866bde29328fc46171db/expertise/models/multifacet_recommender/specter.py#L354) when loading publications, `load_from_redis` is called, but when loading submissions, `load_emb_file` is called.

melisabok: I would prefer to have the caching logic in one single place. Is it possible to check this when the embedding is being computed?:

```
if cache.exists(publication.id):
     embeddings = cache.get(publication.id)
else:
     embeddings = compute_embeddings(publication.id)
     cache.set(publication.id, embeddings)
```

Maybe this is difficult to add giving the current implementation.

tests/test_expertise_service.py
purujitgoyal: I think I did it by mistake, that's why my tests are failing. Will fix this.

expertise/models/multifacet_recommender/specter.py
```diff
@@ -81,6 +87,22 @@ def run(self) -> None:
 
         if self._output_file is not None:
             self._output_file.close()
+        self._redis_con.close()
+
+    def _maybe_print_to_console_and_file(self,
```
melisabok, line 92: who's calling this function?

purujitgoyal: It's called by specter internally while calculating embeddings.

expertise/models/multifacet_recommender/specter.py
melisabok: Yes, but you can create the client outside of this object and then this is easier to test and you don't need to read from a configuration file here to know the redis connection args.

expertise/models/multifacet_recommender/specter.py
melisabok: I see, it is a function defined by specter.

expertise/models/multifacet_recommender/specter.py
purujitgoyal: yes, that can be done.

melisabok:
I had a call with Purujit and he explained to me the implementation. I understand now and I see the difficulties of the current integration with Specter that doesn't allow us the change to be in a single place. 

@carlosmondra @haroldrubio please take a look and leave your comments.

expertise/models/multifacet_recommender/specter.py
```diff
@@ -13,6 +13,9 @@
 import torch
 from tqdm import tqdm
 from typing import Optional
+import redisai
+import redis
```
carlosmondra, line 17: So we need both libraries for redisAI to work?

purujitgoyal: I think only `redisai` should suffice, will have to check.

carlosmondra (approved):
👍

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     working_directory: ~/openreview-expertise
     docker:
       - image: ubuntu:18.04
-      - image: circleci/redis:6.0.0
+      - image: redislabs/redisai
```
haroldrubio, line 8: Does this also start a regular instance of Redis/would there be an issue with keeping both images in CircleCI? I'm wondering because my Redis PR does not use redisai

expertise/service/__init__.py
```diff
@@ -73,3 +77,11 @@ def create_celery(app):
     )
 
     return celery
+
+
+def create_redis(app):
+    """
+    Initializes a redis connection pool
+    """
+    pool = redis.ConnectionPool(host='localhost', port=6379, db=5)
```
haroldrubio, line 86: I think we should store these in the Flask config so we know where all the hardcoded parameters are, here's how I did them in my PR: https://github.com/openreview/openreview-expertise/blob/cb52038ca6adff2430b0aec41945985457822e42/expertise/service/__init__.py#L83-L87

Will we need 2 connection pools since you chose to use db 5 and mine is currently on db 10?

expertise/utils/utils.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import print_function, absolute_import, unicode_literals
 import codecs
+import shutil
```
haroldrubio, line 3: Looks like these changes are leftover from my original code, you can remove them to make the branch cleaner

.circleci/config.yml
purujitgoyal: Yes, `redisai` image launches regular instance of Redis as well, we don't need the circle ci redis.

haroldrubio:
Because the changes are at the model-level, it looks like this call is causing the command-line program to fail:
https://github.com/openreview/openreview-expertise/blob/b3f259fe64b05f99f4bd15c0ff6603fe0fc30b28/expertise/models/multifacet_recommender/specter.py#L183

I think we have 2 approaches here:
1. **Always Cache and Update Documentation**: This would have to be done for the *Running The Server* section anyway. But if we want to always cache, then we have to update the general installation instructions with how to run the RedisAI docker image - and that other users of this repo would have to be running this as well.
2. **Only Cache When Flagged**: We can add a new keyword to the expertise configuration files that flags the program to load/store cached embeddings so as to make sure that other users don't have to run this docker image. This flag would be hardcoded to true in the default `model_params` of the server.

melisabok:
Could you also add a test the calls to the expertise API twice with the same parameters and both outputs are the same?

haroldrubio (approved):

melisabok (approved):

PR #114, opened 2022-05-10
Allow blind publications for expertise

### LINKED ISSUE #112: Deduplicate blind and original submissions
Previously, the expertise CLI was used by the Super User who has access to all data, including the original notes that usually reply to invitations ending with `/-/Submission`. Since the expertise and the matching are now services, we need to get the notes that reply to the `/-/Blind_Submission` invitations. However, the Super User will get duplicates (original and blind submission).

In order to deduplicate the Notes, we can take advantage of the `original` field value in the Blind Notes.

haroldrubio:
- Resolves #112

Removes the check for only original notes and adds an additional deduplication step when retrieving a reviewer's expertise.

### DIFF
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -58,10 +58,10 @@ class OpenReviewExpertise(object):
         deduplicated = []
 
         # Build index of pub.original
-        original_ids = [getattr(pub, 'original', None) for pub in publications]
+        original_ids = [pub.original for pub in publications]
 
         for pub in publications:
-            note_id, original_id = pub.id, getattr(pub, 'original', None)
+            note_id, original_id = pub.id, pub.original
             # If original note, but blind already exists, skip the original note
             if original_id is None and note_id in original_ids:
                 continue
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -57,13 +57,13 @@ class OpenReviewExpertise(object):
     def deduplicate_publications(self, publications):
         deduplicated = []
 
-        # Build index of publication IDs
-        pub_ids = [pub.id for pub in publications]
+        # Build index of pub.original
+        original_ids = [getattr(pub, 'original', None) for pub in publications]
 
         for pub in publications:
-            original = getattr(pub, 'original', None)
-            # If blind note, but the original already exists, skip this
-            if original is not None and original in pub_ids:
+            note_id, original_id = pub.id, getattr(pub, 'original', None)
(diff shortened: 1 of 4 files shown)


expertise/create_dataset.py
```diff
@@ -54,13 +54,32 @@ def get_paper_notes(self, author_id, dataset_params):
         notes_v2 = list(openreview.tools.iterget_notes(self.openreview_client_v2, content={'authorids': author_id}))
         return notes_v1 + notes_v2
 
+    def deduplicate_publications(self, publications):
+        deduplicated = []
+
+        # Build index of pub.original
+        original_ids = [getattr(pub, 'original', None) for pub in publications]
```
melisabok, line 61: what don't you do `pub.original` here?

expertise/create_dataset.py
```diff
@@ -54,13 +54,32 @@ def get_paper_notes(self, author_id, dataset_params):
...
+
+        for pub in publications:
+            note_id, original_id = pub.id, getattr(pub, 'original', None)
+            # If original note, but blind already exists, skip the original note
+            if original_id is None and note_id in original_ids:
+                continue
```
melisabok, line 67: I would prefer to avoid the `continue` sentence and evaluate when we need to keep the publications, my suggestion is:

```
if note not in original_ids:
     deduplicated.append(pub)
```

Only notes that are not original notes should be used as a publication.

expertise/create_dataset.py
melisabok: yeah, this code should be updated. please use `pub.original` from now on. No need to changes other part of the code you are editing now.

melisabok (approved):

expertise/create_dataset.py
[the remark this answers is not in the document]
haroldrubio: Here is what I have now: https://github.com/openreview/openreview-expertise/blob/9501e0eda62958005434ff3daf5f0394aef1ed69/expertise/create_dataset.py#L61

Do you mean to make a change like the following:
`original_ids = [pub.original if pub.original is not None for pub in publications] `
to avoid repeatedly comparing against None values?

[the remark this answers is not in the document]
carlosmondra: Yes, but I also mean that you could use a [set](https://docs.python.org/3/tutorial/datastructures.html#sets). This will make it more efficient when you check for duplicates.

carlosmondra (approved):

PR #120, opened 2022-05-19
Fix for empty title or abstract

purujitgoyal:
Skips the publication before passing to specter if both `title` and `abstract` are empty.


expertise/models/multifacet_recommender/specter.py
```diff
@@ -180,19 +180,22 @@ def set_archives_dataset(self, archives_dataset):
...
-                    else:
-                        paper_ids_list.append(publication['id'])
-                        output_dict[publication['id']] = {"title": self.pub_note_id_to_title[publication['id']],
-                                                          "abstract": self.pub_note_id_to_abstract[publication['id']],
-                                                          "paper_id": publication["id"],
-                                                          "authors": [profile_id]}
+                if publication['content'].get('title') or publication['content'].get('abstract'):
+                    self.pub_note_id_to_author_ids[publication['id']].append(profile_id)
+                    self.pub_author_ids_to_note_id[profile_id].append(publication['id'])
+                    self.pub_note_id_to_title[publication['id']] = publication['content'].get('title') if publication['content'].get('title') else "."
```
melisabok, line 186: why the default value is "." when there is no title? so specter doesn't ignore this publication?

purujitgoyal: Yes, that's correct.

melisabok (approved):

carlosmondra (approved):

PR #122, opened 2022-05-27
Fix for only spaces in Title or abstract

purujitgoyal:


melisabok (approved):
it looks good but the build is failing

haroldrubio (approved):

PR #123, opened 2022-06-02
Use OpenReview API for Testing

haroldrubio:
This PR deprecates the use of a mock client to create the dataset and now requires a locally host OpenReview API to perform the tests


melisabok:
there is a conflict here @haroldrubio

tests/data/api2Data.json
```diff
@@ -18,24 +18,33 @@
         }
       }
     ],
+    "HIJ.cc/-/Blind_Submission": [
```
melisabok, line 21: do you still need to keep these json files?

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+        pc_client=openreview.Client(username='pc@neurips.cc', password='1234')
+        request_form=client.get_notes(invitation='openreview.net/Support/-/Request_Form', sort='tmdate')[0]
+
+        conference=openreview.helpers.get_conference(pc_client, request_form.id)
+        return conference
+
+    def test_create_client(self, client, openreview_client):
+        # Test connectivity to the API
+
+        client = openreview.Client()
```
melisabok, line 24: why do you need to do this? `client` and `openreview_client` should be already created.

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+        openreview_client = openreview.Client()
+        assert openreview_client
+        assert openreview_client.token
+        assert openreview_client.profile
+        assert '~Super_User1' == openreview_client.profile.id
+
+    def test_create_groups(self, client, helpers):
+        # Post test data groups to the API
+
+        def post_profiles(data):
```
melisabok, line 39: instead of creating the profiles, can you use `helpers.create_user`?

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+
+        group = openreview.Group(
+            id = 'ABC.cc',
+            readers = ['everyone'],
+            writers = ['openreview.net'],
+            signatories = ['openreview.net'],
+            signatures = ['openreview.net'],
+            members = members
+        )
+        client.post_group(group)
```
melisabok, line 97: if you are creating a conference, why don't you use `conference.get_reviewers_id()` and add the members to that group. It is a more realistic test.

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+                            }
+                        }
+                }
+            )
+        client.post_invitation(invitation)
+        assert client.get_invitation('DEF.cc/-/Expertise_Selection')
+
+    def test_post_submissions(self, client, openreview_client):
+        
+        def post_notes(data, data_invitation, api_invitation):
```
melisabok, line 281: what's the difference between `data_invitation` and `api_invitation`?

tests/test_create_journal.py
```diff
@@ -0,0 +1,107 @@
...
+                            'supplementary_material': { 'value': '/attachment/' + 's' * 40 +'.zip'},
+                            'competing_interests': { 'value': 'None beyond the authors normal conflict of interests'},
+                            'human_subjects_reporting': { 'value': 'Not applicable'},
+                            'submission_length': { 'value': 'Regular submission (no more than 12 pages of main content)'}
+                        }
+                    ))
+
+        with open('tests/data/api2Data.json') as json_file:
+            data = json.load(json_file)
+        post_notes(data, 'ABC.cc/-/Blind_Submission')
```
melisabok, line 67: the API2 shouldn't have blind submissions...

tests/data/api2Data.json
haroldrubio: Right now, the different files have data for different purposes but I could work on merging all of the files into the single `fakeData.json` file

tests/test_create_conference.py
haroldrubio: The `post_profiles` function calls `helpers.create_user` after reading the profile data from the JSON files

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+                signatures = ['openreview.net'],
+                readers = ['everyone'],
+                invitees = ['everyone'],
+                reply=reply
+            )
+        client.post_invitation(invitation)
+        assert client.get_invitation('DEF.cc/-/Blind_Submission')
+
+        # create_dataset edge invitation to test exclusion
+        invitation = openreview.Invitation(
```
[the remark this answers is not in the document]
haroldrubio: Ah okay, I can use a Conference object which should address everything above as well, I can't remember why I made the decision to make the groups/invitations manually

tests/test_create_conference.py
haroldrubio: The fake notes are stored in the JSON as:
```
{
"notes": {
     "conf_id/-/submission": [ ... ]
}
```
so the distinction is that the fake data stored on disk is formatted differently from the invitation that they will be posted to. The real solution here that I'll work on would be to just store the data exactly as it would be posted

tests/test_create_conference.py
melisabok: but create_user already creates the profile. It calls to the signup endpoint and active the profile.

tests/test_create_conference.py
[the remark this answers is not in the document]
melisabok: you can use the Conference object directly, like the Journal class, or use the request form.

tests/test_create_conference.py
haroldrubio: Right, that happens here:
https://github.com/openreview/openreview-expertise/blob/a2e011a580d294d1714766726b4ba77a73ac0490/tests/test_create_conference.py#L44-L47

Some of the tests reference profile IDs that end in a number greater than 1, so this function checks if that's the case and renames the profile

melisabok: Could you just use all different name so all the tilde ids end with 1?

melisabok: you can use different names for each test or have a single function to be called before all the tests that create all the profiles for all the tests.

expertise/service/celery_tasks.py
```diff
@@ -60,18 +60,14 @@ def after_expertise_return(self, status, retval, task_id, args, kwargs, einfo):
...
-        openreview_client = openreview.Client(
-            token=token,
-            baseurl=config.baseurl
-        )
-        openreview_client_v2 = openreview.api.OpenReviewClient(
-            token=token,
-            baseurl=config.baseurl_v2
-        )
-    else:
-        openreview_client = mock_client(version=1)
```
melisabok, line 73: I like that mock_client is not here anymore!

tests/test_create_conference.py
```diff
@@ -0,0 +1,389 @@
...
+                    email = profile_json.get('content').get('preferredEmail') or profile_json.get('content').get('emails')[0]
+                    first_name = profile_json['id'][1:-1].split('_')[0]
+                    last_name = profile_json['id'][1:-1].split('_')[-1]
+                    helpers.create_user(email, first_name, last_name)
+                    
+                    # A pre-generated user may not have 1 at the end
+                    # In this case, add the same name to their profile, create a group for this new ID and rename the profile
+                    curr_id = client.get_profile(email).id                   
+                    if new_id != curr_id:
+                        client.post_profile(openreview.Profile(
```
melisabok, line 170: do you still need to do this?

tests/test_create_conference.py
```diff
@@ -0,0 +1,339 @@
...
+            signatories = ['openreview.net'],
+            signatures = ['openreview.net'],
+            members = members
+        )
+        client.post_group(group)
+
+    def test_create_invitations(self, client, openreview_client):
+        # Post invitations for submissions and publications
+
+        reply = {
```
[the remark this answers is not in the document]
melisabok: I don't think you need to create this invitation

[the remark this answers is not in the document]
haroldrubio: I kept the archival invitation (openreview.net/-/paper) for previous publications -  do you think that I should just make another conference and use that conference's submission invitation for the test user publications?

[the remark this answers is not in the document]
melisabok: No, that's fine. I'm working on an archive invitation to be used by the tests. You can use it once my PR is merged.

melisabok (dismissed):

.circleci/config.yml
```diff
@@ -1,11 +1,20 @@
...
 
+orbs: 
+  node: circleci/node@5.0.2
 jobs:
   build:
     working_directory: ~/openreview-expertise
     docker:
       - image: ubuntu:18.04
       - image: redislabs/redisai
+      - image: circleci/mongo:4.4.5
```
carlosmondra, line 11: Let's replace this to be mongo v5 😄

haroldrubio: Any version in particular that you had in mind or just replace it with `circleci/mongo:5.0`?

carlosmondra: Any version of 5.0.x should be fine.

.circleci/config.yml
```diff
@@ -50,6 +65,75 @@ jobs:
...
             pip install -I protobuf==3.20.1
+      - run:
+          name: Initialize Replica Set
+          command: |
+            wget -qO - https://www.mongodb.org/static/pgp/server-4.4.asc | sudo apt-key add -
+            sudo touch /etc/apt/sources.list.d/mongodb-org-4.4.list
+            echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu bionic/mongodb-org/4.4 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-4.4.list
+            sudo apt-get update
+            sudo apt-get install -y mongodb-org=4.4.5 mongodb-org-server=4.4.5 mongodb-org-shell=4.4.5 mongodb-org-mongos=4.4.5 mongodb-org-tools=4.4.5
+            mongo mongodb://localhost:27017 --eval "rs.initiate()"
```
carlosmondra, line 76: You will need to update this too, but it should be pretty much the same thing

carlosmondra (approved):
Awesome!

PR #127, opened 2022-08-31
Feature/Add Inclusion Invitation

haroldrubio:
This PR adds support for the key 'include_expertise_selection'. This makes it so the `/-/Expertise_Selection` invitation posts edges with label 'Include' instead of 'Exclude'. Users with these edges choose which papers count for their expertise, instead of excluding specific papers from their overall publication history.

This PR also adds additional descriptors to common errors that occur when a venue has no submissions and/or when the specified match group is empty


tests/test_create_conference.py
```diff
@@ -247,6 +247,61 @@ def test_create_invitations(self, client, openreview_client):
...
+                "label": {
+                    "value-radio": [
+                        "Include"
+                    ],
+                    "required": True
+                }
+            }
+        }
+
+        invitation = openreview.Invitation(
```
melisabok, line 283: instead of creating the invitation, could you set the flag in the conference object? 

https://github.com/openreview/openreview-py/blob/master/openreview/venue_request/venue_request.py#L1139

'Yes' means the inclusion is on, otherwise the invitation os being use to exclude. Be aware that we use the same invitation id in both cases.

expertise/service/utils.py
```diff
@@ -311,37 +314,54 @@ def _camel_to_snake(camel_str):
...
 
         if api_request.entityA['type'] == 'Group':
             config.match_group = [api_request.entityA['memberOf']]
+            edge_inv = api_request.entityA.get('expertise', None)
+
+            if edge_inv:
+                edge_inv_id = edge_inv.get('invitation', None)
+                if edge_inv_id is None or len(edge_inv_id) <= 0:
+                    raise openreview.OpenReviewException('Bad request: Expertise invitation indicated but ID not provided')
+                if 'Inclusion' in edge_inv_id:
```
melisabok, line 328: please see my previous comment, the invitation id doesn't define if it is an inclusion or exclusion. I believe the best way to do it is inspecting the label options.

expertise/service/utils.py
```diff
@@ -311,37 +314,58 @@ def _camel_to_snake(camel_str):
...
             config.match_group = [api_request.entityA['memberOf']]
+            edge_inv = api_request.entityA.get('expertise', None)
+
+            if edge_inv:
+                edge_inv_id = edge_inv.get('invitation', None)
+                if edge_inv_id is None or len(edge_inv_id) <= 0:
+                    raise openreview.OpenReviewException('Bad request: Expertise invitation indicated but ID not provided')
+                label = openreview_client.get_invitation(edge_inv_id).reply['content']['label']['value-radio'][0]
+                print(f"FETCHED LABEL")
+                if 'exclude' not in label.lower():
```
melisabok, line 330:
```suggestion
                if 'include' in label.lower():
```

expertise/service/utils.py
```diff
@@ -311,37 +314,58 @@ def _camel_to_snake(camel_str):
...
             config.alternate_match_group = [api_request.entityB['memberOf']]
+            edge_inv = api_request.entityB.get('expertise', None)
+
+            if edge_inv:
+                edge_inv_id = edge_inv.get('invitation', None)
+                if edge_inv_id is None:
+                    raise openreview.OpenReviewException('Bad request: Expertise invitation indicated but ID not provided')
+                label = openreview_client.get_invitation(edge_inv_id).reply['content']['label']['value-radio'][0]
+                print(f"FETCHED LABEL")
+                if 'exclude' not in label.lower():
```
melisabok, line 345:
```suggestion
                if 'include' in label.lower():
```

expertise/service/utils.py
```diff
@@ -311,37 +314,58 @@ def _camel_to_snake(camel_str):
...
+
         if api_request.entityB['type'] == 'Group':
             config.alternate_match_group = [api_request.entityB['memberOf']]
+            edge_inv = api_request.entityB.get('expertise', None)
+
+            if edge_inv:
+                edge_inv_id = edge_inv.get('invitation', None)
+                if edge_inv_id is None:
+                    raise openreview.OpenReviewException('Bad request: Expertise invitation indicated but ID not provided')
+                label = openreview_client.get_invitation(edge_inv_id).reply['content']['label']['value-radio'][0]
```
melisabok, line 343: when we give support to the new API, the invitation scheme will be different, like:

```
invitation.edge['content']['label']['param']['enum']
```

I also would try to use get() to avoid the KeyError exception. You could validate the format of the invitation and throw an error if it is not correct.

melisabok:
looks good to me, @carlosmondra please take a look.

carlosmondra:
What is this PR supposed to do?

melisabok (dismissed):

melisabok:
> What is this PR supposed to do?

@carlosmondra please check the description of the PR if something is not clear, @haroldrubio please add more info!

expertise/service/utils.py
```diff
@@ -311,37 +314,56 @@ def _camel_to_snake(camel_str):
...
+
+            if edge_inv:
+                edge_inv_id = edge_inv.get('invitation', None)
+                if edge_inv_id is None:
+                    raise openreview.OpenReviewException('Bad request: Expertise invitation indicated but ID not provided')
+                label = openreview_client.get_invitation(edge_inv_id).reply.get('content', {}).get('label', {}).get('value-radio',['Include'])[0]
+                if 'include' in label.lower():
+                    config.inclusion_inv = edge_inv_id
+                else:
+                    config.exclusion_inv = edge_inv_id
```
carlosmondra, line 346: You could place the repeated code in a function

melisabok (approved):

PR #129, opened 2022-09-20
Fix/group expertise selection

### LINKED ISSUE #124: Group scoring does not take into account expertise selection
When retrieving all publications as submissions, we only use iterget with no checking of the paperhash and expertise selection. Make the same checks that happen when retrieving publications as expertise.

haroldrubio:
- Resolves #124.
- Requires #127.

This PR introduces 2 new keywords into the config file for `create_dataset`: `alternate_exclusion_inv` and `alternate_inclusion_inv` which allows the alternate match group to have their expertise taken into account when performing group-group scoring.


haroldrubio:
Merged into:
- #127

PR #136, opened 2022-11-10
Fix Specter cache

haroldrubio:
This PR no longer attempts to find and remove all old instances of publications in the Redis cache, and instead sets an expiration date whenever inserting into Redis.

### DIFF
--- a/expertise/models/multifacet_recommender/specter.py
+++ b/expertise/models/multifacet_recommender/specter.py
@@ -116,6 +116,7 @@ class _PredictManagerCustom(_PredictManager):
             paper_id = prediction_json['paper_id']
             cache_key = paper_id + "_" + str(self._metadata[paper_id]['mdate'])
             self._redis_con.tensorset(key=cache_key, tensor=np.array(prediction_json['embedding']))
+            self._redis_con.expire(cache_key, 2629746) ## Expire after 1 month
 
 
 def predictor_from_archive(archive: Archive, predictor_name: str = None,
--- a/expertise/models/multifacet_recommender/specter.py
+++ b/expertise/models/multifacet_recommender/specter.py
@@ -211,7 +211,6 @@ class SpecterPredictor:
                                 "authors": [profile_id],
                                 "mdate": pub_mdate
                             }
-                        self._remove_keys_from_cache(publication["id"])
                 else:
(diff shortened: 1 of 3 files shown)


expertise/models/multifacet_recommender/specter.py
```diff
@@ -211,7 +212,6 @@ def set_archives_dataset(self, archives_dataset):
                                 "authors": [profile_id],
                                 "mdate": pub_mdate
                             }
-                        self._remove_keys_from_cache(publication["id"])
```
melisabok, line 214: so this is the call that is taking time? `tensorset` is fast enough?

haroldrubio: Yes, tensorset is fast but the `_remove_keys_from_cache` function scans through all the keys to find matches which is quite slow

expertise/models/multifacet_recommender/specter.py
```diff
@@ -116,6 +116,7 @@ def _maybe_print_to_console_and_file(self,
             paper_id = prediction_json['paper_id']
             cache_key = paper_id + "_" + str(self._metadata[paper_id]['mdate'])
             self._redis_con.tensorset(key=cache_key, tensor=np.array(prediction_json['embedding']))
+            self._redis_con.expire(cache_key, 2629746) ## Expire after 1 month
```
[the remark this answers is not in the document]
haroldrubio: Hmm it can be set in the `model_params` of the `config.json` but I don't think the models will have access to the Flask config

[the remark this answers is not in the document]
melisabok: ok, let's keep it as it is

expertise/models/multifacet_recommender/specter.py
carlosmondra: How much is slow? I'm still confused as to why redis is slow here. Are there too many keys in the database?

I'm not sure what the `self._remove_keys_from_cache` method does, but its performance also depends on whether [`SCAN` or `KEYS`](https://stackoverflow.com/questions/32603964/scan-vs-keys-performance-in-redis) is used.

haroldrubio:
No longer using RedisAI cache - embedding cache will be implemented in Google Cloud

PR #137, opened 2022-11-10
Cancels jobs that were running on server restart

haroldrubio:
This PR adds additional startup code to Flask and Celery that marks all jobs running at the time of a server restart with ERROR which will cause process functions to fail while clearing the Celery queue.

### DIFF
--- a/expertise/service/__init__.py
+++ b/expertise/service/__init__.py
@@ -2,8 +2,10 @@ import os
 import flask
 import logging, logging.handlers
 import redis
+import time
 
 from celery import Celery
+from expertise.service.utils import JobStatus, JobDescription, RedisDatabase
 
 def configure_logger(app):
     '''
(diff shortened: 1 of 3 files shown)


expertise/service/__init__.py
```diff
@@ -62,20 +65,94 @@ def create_app(config=None):
...
+    def set_config_error(job_config):
+        status = job_config.status
+        if status in [
+            JobStatus.QUEUED,
+            JobStatus.FETCHING_DATA,
+            JobStatus.EXPERTISE_QUEUED,
+            JobStatus.RUN_EXPERTISE
+        ]:
+            app.logger.info(f"{job_config.job_id} was running - canceling job")
+            job_config.status = JobStatus.ERROR
```
melisabok, line 104: Shouldn't the status be "Cancelled"? I would like to distinguish it from error.

carlosmondra: Does the "canceled" status make the process functions fail and clears the celery queue?

haroldrubio: I was thinking about this - I did create a Cancel status but openreview-py will only fail if it sees an "Error", otherwise the process function will wait a day before it times out. I can make it canceled but we would need to update openreview-py to catch the "Cancel" status.

melisabok:
should we close this PR too?

haroldrubio:
Not applicable - no longer using Celery

PR #138, opened 2022-11-10
Clean jobs script

haroldrubio:
This script will be set on a cron job and is used to remove job data except for the config file and score file(s) for jobs older than a month.


melisabok (approved):

carlosmondra (approved):
Nice solution. So files/directories will be kept for 30 days right?

haroldrubio:
Yes, and I was thinking about running a cronjob maybe every day?

PR #140, opened 2022-11-30
Check edge label

### LINKED ISSUE #139: Create dataset: when reading the expertise edges we should check if the label matched with "Include" or "Exclude"
Some edges could have been created by mistake, we need to only read the edges that matches with the invitation:

https://github.com/openreview/openreview-expertise/blob/master/expertise/create_dataset.py#L291

haroldrubio:
- Resolves #139 

This PR slightly refactors the code for getting expertise selection edges and only adds edges whose labels match with the invitation

### DIFF
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -245,7 +245,7 @@ class OpenReviewExpertise(object):
                 self.openreview_client,
                 invitation=invitation,
                 groupby='tail',
-                select='id,head,label,weight,label'
+                select='id,head,label,weight'
             )
 
             for edges in user_grouped_edges:
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -256,72 +256,16 @@ class OpenReviewExpertise(object):
         return selected_ids_by_user
 
     def exclude(self):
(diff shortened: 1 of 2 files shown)


expertise/create_dataset.py
```diff
@@ -237,73 +237,35 @@ def get_profile_ids(self, group_ids=None, reviewer_ids=None):
...
+    def get_expertise_selection_edges(self, invitation_key, label=None):
+        edge_invitations = self.convert_to_list(self.config[invitation_key])
+        selected_ids_by_user = defaultdict(list)
+        for invitation in edge_invitations:
             user_grouped_edges = openreview.tools.iterget_grouped_edges(
                 self.openreview_client,
                 invitation=invitation,
                 groupby='tail',
-                select='id,head,label,weight'
+                select='id,head,label,weight,label'
```
melisabok, line 248: label seems to be duplicated here.

melisabok: adding the parameter `label=` to iterget_grouped_edges` doesn't work?

melisabok (approved):

PR #151, opened 2023-05-03
Fix tests

haroldrubio:
This PR adds/removes some fields from the venue request forms that were changed in openreview-py to get the tests to pass


melisabok (approved):
thanks for fixing the tests!

PR #154, opened 2023-05-12
Match /status query params with definition

haroldrubio:
This PR now has the `/status` endpoint check for `id` (and `entityA/B.id`) and `invitation` (and `entityA/B.invitation`) instead of `paperId` and `paperInvitation` in the query params


melisabok (approved):
thanks!

PR #167, opened 2024-01-17
Fix tests: add submission license field

melisabok:

PR #168, opened 2024-01-18
Fix CircleCI Tests with New Link

haroldrubio:
This PR no longer downloads from Google Drive and instead downloads from a file hosted in an OpenReview bucket


haroldrubio:
Closing since the tests are still going to fail because the venue requests are missing the submission fields. Moving changes to the submission license PR

- #167

PR #170, opened 2024-02-08
Remove Redis Cache From Tests

haroldrubio:
This PR removes the Redis cache from the tests as it is no longer being used - we can fully deprecate it in the models when we find another way to cache the embeddings


carlosmondra (approved):
Thanks!

Issue #166, opened 2024-01-02
Expertise API: Change Delete Endpoint


haroldrubio:
The delete endpoint should be changed to a POST operation

There is also some issue with backwards compatibility when returning the delete config

Issue #156, opened 2023-06-15
Statuses are not properly set


haroldrubio:
There is a gap in the way the job statuses are set in celery tasks which makes the statuses returned by the API to not be completely accurate.

haroldrubio:
The task IDs are also not set so there is no synchronization when a job enters the `expertise` queue, it gets a random task ID.

haroldrubio:
Resolved with migration to Google Cloud Vertex AI

PR #174, opened 2024-02-14
Improvements to Delete Endpoint and Statuses

haroldrubio:
Resolves
- #166 
- #156 

This PR contains the following changes:
- Switches the `/delete` endpoint to POST
- Adds a new status `Revoked` to the JobConfig
- Fixes a bug where jobs that were waiting to have their scores computed were displayed as running, now they are shown as queued for expertise
- Adds the JobConfig as a dict to the Celery task arguments so the job information is visible when querying the queue by the Celery CLI
- Changes the logic of the `/delete` endpoint:
    - If called on a running job (any job whose status is not error or completed), sets the status to revoked, the job maintains the revoked status and gets cleaned up after making a final check in the `expertise` Celery task
    - If called on a non-running jobs (either error or completed), cleans the files on disk and removes the job from Redis
- Changes the high load test (which queues 5 jobs) to revoke the last job and checks that it was cleaned up correctly

### DIFF
--- a/expertise/service/celery_tasks.py
+++ b/expertise/service/celery_tasks.py
@@ -34,6 +34,19 @@ def update_status(config, new_status, desc=None):
     redis_db = RedisDatabase(connection_pool=redis_config_pool)
     redis_db.save_job(config)
 
+def check_revoked(config: JobConfig):
+    """
+    Gets the status given a config - used to check for revoked status
+
+    :param config: JobConfig of a given job
+    :type config: JobConfig
+
+    :return True if revoked:
+    """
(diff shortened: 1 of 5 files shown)


melisabok:
the changes look good!

I'm wondering if this PR is still active: https://github.com/openreview/openreview-expertise/pull/137

I think the jobs that were running and the server is restarted should be change to status "revoked", right?

melisabok:
I think this PR can be closed given that we don't use Celery anymore, @haroldrubio can you please confirm?

PR #181, opened 2024-02-26
Use PyTorch for New Model Embeddings

haroldrubio:
This PR moves more shared functions into the Predictor class, avoids moving the embeddings from GPU to CPU each batch (speeds up each iteration), uses PyTorch's `.save()` and `.load()` to store embeddings more efficiently (takes up less disk)

### DIFF
--- a/expertise/models/specter2_scincl/predictor.py
+++ b/expertise/models/specter2_scincl/predictor.py
@@ -26,3 +26,26 @@ class Predictor:
             if not batch:
                 break
             yield batch
+
+    def _batch_predict(self, batch_data):
+        jsonl_out = []
+        text_batch = [d[1]['title'] + self.tokenizer.sep_token + (d[1].get('abstract') or '') for d in batch_data]
+        # preprocess the input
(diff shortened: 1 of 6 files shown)


expertise/models/specter2_scincl/predictor.py
```diff
@@ -1,4 +1,7 @@
 from tqdm import tqdm
+import openreview
```
melisabok, line 2: why do you need to include openreview? just for triggering the exception? maybe this can be a regular exception

melisabok:
it seems this only works for specter2+scinclr, can we support it for specter2+mfr too?

haroldrubio:
It looks like MFR is already using PyTorch serialization. The original SPECTER would be pretty tough to override since it looks like it re-uses some old code from a 4 year old branch of the allennlp library, the JSON predictions are built into it

carlosmondra:
Do we still want to merge this?

melisabok:
@haroldrubio what should we do about this?

melisabok:
closing this PR, no feedback since March.

PR #191, opened 2024-09-18
Fix: Environment Dependencies

haroldrubio:
This PR fixes issues in the dependencies for the environment by updating the Python version to 3.9, pinning down adapter transformers and spacy and removing the dependency on allennlp.

Deprecates ELMo model as it depends on an incompatible version of allennlp


carlosmondra:
I still found some ELMo strings in the README. Is there a reason to keep those?

carlosmondra (approved):
GREAT!

PR #192, opened 2024-11-14
Fix: await queue function

haroldrubio:
This PR fixes the tests by changing how we await the process logs


carlosmondra (approved):
Thank you!

PR #194, opened 2024-12-11
BullMQ queue

carlosmondra:
This PR:

- Changes the celery queue for the BullMQ queue.
- Adds better error logging which includes the Traceback of the error.
- Ensures that no job with the same entity A, entity B and model is run at the same time.


expertise/service/expertise.py
```diff
@@ -141,29 +177,248 @@ def _get_score_and_metadata_dir(self, search_dir, group_scoring=False):
...
+        if config.paper_id:
+            log.append(f"Paper ID: {config.paper_id}")
+        if config.paper_invitation:
+            log.append(f"Paper invitation: {config.paper_invitation}")
+        if config.paper_venueid:
+            log.append(f"Paper venue ID: {config.paper_venueid}")
+        if config.model:
+            log.append(f"Model: {config.model}")
+        if config.model_params:
+            log.append(f"Model params: {json.dumps(config.to_json().get('model_params', {}), indent=4)}")
```
haroldrubio, line 277: One quick thing: we probably also want to log the `match_group` and `alternate_match_group` (if it exists)

haroldrubio (approved):

PR #196, opened 2024-12-12
Fix: Add headers to network calls

haroldrubio:
This PR adds the openreview_client headers to the test client when making requests to Flask


haroldrubio:
Changes included in #194

PR #198, opened 2024-12-18
Queue: Cleanup resources at the end of each job

haroldrubio:
This PR addresses the issue of processes holding on to GPU memory causing jobs to run out of memory after some time.

This PR explicitly calls the garbage collector and frees the CUDA cache. This does not free **all** of the memory, the process still holds on to some GPU memory for its context, which amounts to 100-200MB per process

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -3,6 +3,8 @@ import shutil
 import time
 import os
 import json
+import torch
+import gc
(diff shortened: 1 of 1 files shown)


haroldrubio:
A more complete solution might be to run the model in its own process so that it gets completely torn down but I don't know if that's best practice

PR #211, opened 2025-01-17
Cloud Service: Expertise status returns only jobs in Redis

haroldrubio:
This PR prevents jobs that aren't in Redis but are in the bucket (meaning, there is no `jobId` substitution so the return value will be incompatible with the expertise API) from being returned from the /status endpoint when not providing a `jobId` query param


melisabok:
can we just use the same jobId everywhere?

haroldrubio:
Fixed in #215

PR #215, opened 2025-01-28
Cloud Service: Query Redis for All Status

haroldrubio:
This PR changes the behavior of the server when searching for jobs by parameters and not by `jobId`, by avoiding calls to the bucket

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -182,7 +182,7 @@ class BaseExpertiseService:
         if not self.containerized:
             self.redis.save_job(config)
 
-    def _search_redis_jobs(self, user_id, query_params):
+    def get_expertise_all_status(self, user_id, query_params):
         """
         Searches the server for all jobs submitted by a user that satisfies
         the HTTP GET query parameters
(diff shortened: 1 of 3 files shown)


carlosmondra:
Thanks, Harold :)

Here are my comments:

- You created the method `_search_redis_jobs` in the parent class. I was wondering why you didn't just create the method `get_expertise_all_status` in the parent class instead.
- Jobs are pickled and then saved. What is the reason for this? Wasn't `json.dumps` a better alternative? I don't have a strong opinion, just curious.
- The method for getting the job status is very inefficient and to be honest a bit hard to follow the logic. I am not suggesting to fix that in this PR, but I think it would be best if we had a better way for querying jobs. Redis has Redis JSON and Redis search modules. We could include those to store and retrieve jobs more easily. I can create an issue for this so that jobs are queried in a similar fashion to what we do in mongodb.
- How long are jobs stored in redis? I see that there are close to 6000 jobs which isn't much, but we need to make sure it doesn't crash Redis eventually.

haroldrubio:
> Jobs are pickled and then saved. What is the reason for this? Wasn't json.dumps a better alternative? I don't have a strong opinion, just curious.

I think previously, and it still might be the case, that there are objects being stored that aren't JSON serializable.

> The method for getting the job status is very inefficient and to be honest a bit hard to follow the logic. 

Yeah I think that would be great, Redis is very fast but I'm a bit nervous about the scaling in regards to venues with many jobs like TMLR

> How long are jobs stored in redis?

I don't think we have this value set anywhere, or any process to clean up these jobs

Thanks for all the input Carlos! The expertise API wasn't originally planned with this much load in mind so any other feedback on how to build more scalable things is very appreciated

carlosmondra:
Thanks for replying to my comments :). I have created a couple of issues to address them.

PR #219, opened 2025-03-03
Use public papers for expertise

enrubio:
This PR filters publication notes for `readers = everyone`. This will ensure we are using public papers for score computation.

In a new PR we will add the filter for `pdate` and make the same change for the Expertise Selection UI.

### DIFF
--- a/expertise/create_dataset.py
+++ b/expertise/create_dataset.py
@@ -68,9 +68,9 @@ class OpenReviewExpertise(object):
             return [n for n in self.openreview_client.get_notes_by_ids(ids=note_ids) if n.invitation == paper_invitation]
 
         notes_v1 = list(openreview.tools.iterget_notes(self.openreview_client, content={'authorids': author_id}))
-        notes_v1 = [note for note in notes_v1 if note.readers == ['everyone'] and  getattr(note, 'pdate', None)]
+        notes_v1 = [note for note in notes_v1 if note.readers == ['everyone']]
         notes_v2 = list(openreview.tools.iterget_notes(self.openreview_client_v2, content={'authorids': author_id}))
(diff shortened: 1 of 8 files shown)


melisabok (approved):
Thanks Emilia!

melisabok:
I believe you need to fix other tests

expertise/create_dataset.py
```diff
@@ -68,7 +68,9 @@ def get_paper_notes(self, author_id, dataset_params):
             return [n for n in self.openreview_client.get_notes_by_ids(ids=note_ids) if n.invitation == paper_invitation]
 
         notes_v1 = list(openreview.tools.iterget_notes(self.openreview_client, content={'authorids': author_id}))
+        notes_v1 = [note for note in notes_v1 if note.readers == ['everyone'] and  getattr(note, 'pdate', None)]
```
melisabok, line 71: do you all v1 notes have a pdate property @carlosmondra ? we might try removing them. 

Also if we are filtering out by pdate then the Expertise UI should do that same too. @enrubio can you make that change?

carlosmondra: > do you all v1 notes have a pdate property @carlosmondra ? we might try removing them.

No, not all notes in v1 have pdate even if they should. Notably imported v1 dblp notes may not have it. I wonder if we should migrate the v1 dblp notes to v2.

melisabok: > I wonder if we should migrate the v1 dblp notes to v2.
YEs!

melisabok: so we should remove `getattr(note, 'pdate', None)`?

carlosmondra (approved):

haroldrubio (approved):

PR #221, opened 2025-03-09
Feature: Support More Score Aggregation Functions

haroldrubio:
Depends on #220.

This PR exposes `merge_alpha`, `top_k`, `attn_clustering`, and `cross_attention` as arguments to the SPECTER2 and SCINCL models and ensemble model.

The new `top_k` variable will average the highest K paper-paper scores when computing a reviewer's publication scores to a submission. When `top_k` is set to 1, we recover the `max_score: True` setting. When `top_k` is set to 0, we recover the `average_score: True` setting.

Introduces a new module in the SPECTER2 and SciNCL models called `SelfAttentionClusteringPredictor`. It attempts to address the issue that using while max pooling over the publication-submission scores performs best experimentally, it's sensitive to publications in the distant past, one-off collaborations, and gaming the system by deliberately uploading a generated submission.

When `attn_clustering` is true and `cross_attention` is false, the models perform the following:
1. (In batches) Publication cosine similarity self attention scores are computed and summed across the rows. Publications with the highest similarity to all other publications will have the highest scores
2. Normalize the scores and then select the top K scores, and normalize again. If no K is set, all scores are used.
3. Compute a reviewer embedding as the weighted some of their publication embeddings and their scores. The reviewer embedding is now a weighted sum of the most representative embeddings for their expertise.
4. Scores are computed by cosine similarity of the reviewer embeddings to the submission embeddings

When `attn_clustering` is true and `cross_attention` is true, we add more weight to the submission by:
1. (In batches) As before, publication cosine similarity self attention scores are computed and summed across the rows. Publications with the highest similarity to all other publications will have the highest scores.
2. We compute paper-paper scores as before in the original model, between the publications and the submissions
3. Separately, we multiply the self-attention scores with the paper-paper scores (the cross-attention part) and normalize again. This creates weights across the publications that are maximized when a publication is both representative of where most of the reviewer's expertise is, as well as most similar to the submission
4. We select the top K scores and normalize again to filter out outliers to get our final weights. By filter out, we set the non top K scores to a large negative number so their final weights are 0
5. We transform the original paper-paper scores into scores that are weighted by high representation and high similarity, and take the average.

`merge_alpha` also gets exposed as this new attention-based clustering could drastically change how the ensemble model behaves so it would be useful to be able to tune/set this manually


melisabok:
@haroldrubio should we review this PR?

haroldrubio:
Closing because of no intention to merge - renaming with `exp` prefix to indicate an experiment

PR #223, opened 2025-03-10
Expertise API: Fix Singleton Condition

haroldrubio:
This PR fixes an issue where the `user_id` stored in Redis and in the queue does not match the `user_id` stored in GCS. The issue is most likely the sharing of the `ExpertiseCloudInstance` in `routes.py` where we are overwriting the clients per request.

This PR addresses that by explicitly passing the user_id from the request to the `create_job` function

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -792,7 +792,7 @@ class ExpertiseCloudService(BaseExpertiseService):
         request = job.data['request']
         redis_id = job.data['redis_id']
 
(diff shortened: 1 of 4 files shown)


carlosmondra (approved):
Good catch!

haroldrubio:
Made some additional changes to isolate the clients
- Pass the clients from routes to the `start_expertise` function
- Clients get passed to the function that builds the Redis/model config
- Token gets passed to the worker process which logs in a new client and passes that to the Vertex AI job creator

PR #225, opened 2025-03-12
Feature: Support paper-paper scoring

haroldrubio:
This PR adds support for computing affinity scores between notes: by invitation, venue ID or content. The variable/config naming convention was decided because `match_group` spawns the data in the `archives` directory, so any variables that populate the `archives` directory would be prefixed with `match_`.

**Models**
- Add variable `compute_paper_paper` to tell the models to directly write the un-normalized paper-paper scores to the CSV. The unnormalized scores are outputted because we're concerned more with a direct match from the model more than the interpretability
- Adds support to SPECTER, SPECTER2, SciNCL, and SPECTER2+SciNCL

**Package Changes**
- Support `match_paper_invitation`, `match_paper_id`, `match_paper_venueid`, and `match_paper_content` in the config file
- Move the `get_submissions` logic to a helper function to share the code between getting submissions and the submissions that go in the `archives`
- Expose `compute_paper_paper` in the model configs

**API Changes**
- Detect paper-paper scoring by checking if both entities are of type `Note` and populate the `match_*`
- Return non-sparse scores as a list of dicts keyed by `match_submission`, `submission`, and `score`


melisabok:
Thanks for doing this Harold!
 
I understand that we use the token of the API request to get the submissions for each entity. It will be very common that the user is reader of the submissions of one entity but not reader of the other one. What happens in this case? do we return zero scores?

We need to discuss how we can give PCs access to submissions of another venue. For now, we would need to use the super user.

haroldrubio:
Ah that's a good case to test. I can add this, and I would expect that it would throw the same errors as we throw when there are no submissions/publications. But, the error messages in this case should be updated to mention something about paper-paper scoring.

melisabok (approved):

carlosmondra (approved):
I see some code that is repeated between classes. I wonder if it would be worth creating a parent class or interfaces that implement some of the methods to avoid code repetition. It can be done in a separate PR 😬

enrubio (approved):
Cool feature, thank you!

PR #227, opened 2025-03-19
Feat: Implement retry with backoff

carlosmondra:
This PR introduces automatic retries when an expertise job fails


haroldrubio (approved):
Looks good to me!

PR #228, opened 2025-03-20
Feature: CircleCI Building and Executing on Docker Image

haroldrubio:
This PR adds an extra CircleCI job in parallel to the PyTests that:

1. Tests that the image builds successfully and is properly compiled with CUDA
2. Runs a subset of `openreview-expertise` PyTests that checks that the environment is properly setup

TODOs:
- [x] Use tagging to build and push images and Kubeflow pipelines with each release and push to both the version tag and `latest` tag
- [x] Pass in an `openreview-py` version when building the image

To update the pipeline tag in the server, we can add this if statement in the deploy script in the instance:
```
if [ -n "${GCP_PIPELINE_TAG}" ]; then
  echo "Using pipeline tag: ${GCP_PIPELINE_TAG}"
  sed -i "s/^GCP_PIPELINE_TAG=.*/GCP_PIPELINE_TAG=${GCP_PIPELINE_TAG}/" /home/openreview/openreview-expertise/service/config/production.cfg
fi
```


.circleci/config.yml
```diff
@@ -159,3 +159,334 @@ jobs:
...
+            docker run --network host \
+              -e GOOGLE_CLOUD_PROJECT=test-project \
+              -e STORAGE_EMULATOR_HOST=http://localhost:4443 \
+              -e PYTHONUNBUFFERED=1 \
+              -e DEBUG=1 \
+              -e LOG_LEVEL=DEBUG \
+              -v $(pwd)/test_output:/app/output \
+              --entrypoint="" \
+              expertise-test:latest \
+              python -m expertise.execute_pipeline "$(cat test_input.json)"
```
carlosmondra, line 397: Is this a test? Does it mean we can't test this using pytest?

haroldrubio: Yes this is a test - we do have corresponding tests in the `test_pipeline.py` file, but the point of these tests in CircleCI is to make sure that they can run properly in the image - it's as close as we can get to testing the Vertex AI Pipeline

haroldrubio:
I think this PR is looking pretty ready, but I'll need to make the following changes in the deploy script to support deployment by tags:
1) To accept a tag argument in the script
```
TAG=${1:-latest}
echo "Deploying $TAG of openreview-expertise"
```
2) Checkout a specific tag, or master if using "latest"
```
echo "Updating repository to $TAG"
git fetch --all --tags
if [[ $TAG == "master" ]]; then
    git checkout master
    git pull
else
    # For tags, checkout the specific tag
    git checkout "tags/$TAG"
fi
```
3) Update the flask variable `GCP_PIPELINE_TAG` to the newest release:
```
echo "Updating GCP_PIPELINE_TAG in production.cfg"
CONFIG_FILE="/home/openreview/openreview-expertise/expertise/service/config/production.cfg"
if [[ -f "$CONFIG_FILE" ]]; then
    sed -i "s/GCP_PIPELINE_TAG='[^']*'/GCP_PIPELINE_TAG='$TAG'/g" "$CONFIG_FILE"
    echo "Updated GCP_PIPELINE_TAG to '$TAG' in production.cfg"
else
    echo "Warning: production.cfg not found at $CONFIG_FILE"
fi
```

After this, then I can update the deploy action to pass the tag as an argument:
```
VERSION="${{ github.event.inputs.version }}"
echo "Deploying version: $VERSION"
gcloud compute ssh openreview@openreview-expertise-instance --command ".github/workflows/expertise_deploy.sh $VERSION" --quiet
```

haroldrubio:
> What's the difference between push-image and deploy-prod? I suppose the first one builds and pushes the docker image to the artifact registry so that Vertex AI can use it and the second one deploys the latest changes to the VM instance. Shouldn't they be in the same script?

I think deploy-prod could do the same thing as push-image. I can do the same steps as push-image, then call the deployment script on the server. The main thing here is that push-image can be used to push dev images and pipelines to `expertise-test` and `openreview-expertise-test`, so it's more general.

> Also, how do you pass the version you want to deploy to the script in the VM instance?

I just made a commit to actually pass the version argument for the GitHub action as an argument to the deploy script

carlosmondra:
> I think deploy-prod could do the same thing as push-image. I can do the same steps as push-image, then call the deployment script on the server. The main thing here is that push-image can be used to push dev images and pipelines to expertise-test and openreview-expertise-test, so it's more general.

In that case, wouldn't it be better to follow the same approach as the other repos, where we have deploy-prod and deploy-staging?

haroldrubio:
> In that case, wouldn't it be better to follow the same approach as the other repos, where we have deploy-prod and deploy-staging?

That's true, I guess this would mean setting up Vertex AI for the dev site as well?

carlosmondra:
> > In that case, wouldn't it be better to follow the same approach as the other repos, where we have deploy-prod and deploy-staging?
> 
> That's true, I guess this would mean setting up Vertex AI for the dev site as well?

I thought that this was already the case. What would expertise-test be used for then?

haroldrubio:
I've been using the `expertise-test` name for images that I've been manually pushing and testing, the dev site could use something like `expertise-dev`

carlosmondra:
> I've added it to the deploy action, should I remove the standalone action to build and push?

Yeah, I don't think it's needed

> One more thing: what version of openreview-py should I build within the image? Right now it's getting built with master but should I try to fetch the latest tag? Or should I manually ask for the tag of openreview-py to deploy each time we run the script?

You can use the openreview-py version that is deployed to the API v2. You can get the tag that is deployed in production from this file: gs://openreview-general/conf/deployment.conf

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM nvidia/cuda:12.6.3-cudnn-runtime-ubuntu24.04
 
+ARG OPENREVIEW_PY_VERSION=latest
```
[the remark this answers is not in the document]
haroldrubio: The value of this argument gets set here:
https://github.com/openreview/openreview-expertise/blob/5873a9f14d3dc052f0d7c8d2358d329766c1a940/.github/workflows/deploy-prod.yml#L95-L96

I could set the default to `master` instead to remove the need for this if statement:
https://github.com/openreview/openreview-expertise/blob/5873a9f14d3dc052f0d7c8d2358d329766c1a940/Dockerfile#L41-L45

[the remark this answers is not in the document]
carlosmondra: OK, I understand. Yes, let's use master to remove the if statement 👍

.github/workflows/deploy-prod.yml
```diff
@@ -24,16 +39,77 @@ jobs:
...
+          
+          if docker manifest inspect "${IMAGE_PATH}" > /dev/null 2>&1; then
+            echo "exists=true" >> $GITHUB_OUTPUT
+          else
+            echo "exists=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
```
carlosmondra, line 78: Does it matter what version of python we use? I think we used 3.11 in other parts of the PR.

haroldrubio: I don't think it matters since we're only building the pipeline, but I agree to use 3.11 for consistency

PR #229, opened 2025-03-20
Expertise Cloud Service: Paper paper fixes

haroldrubio:
This PR fixes some issues with naming folders in GCS when doing paper-paper matching and when fetching scores.

It also makes a small change to ensure CUDA is installed

### DIFF
--- a/expertise/service/utils.py
+++ b/expertise/service/utils.py
@@ -692,31 +692,53 @@ class GCPInterface(object):
         self.client = client
 
     def _generate_vertex_prefix(api_request):
-        # Precondition: api_request there is at least 1 group entity containing a memberOf field
         group_entity = None
         if api_request.entityA['type'] == 'Group':
             group_entity = api_request.entityA
         elif api_request.entityB['type'] == 'Group':
             group_entity = api_request.entityB
 
-        if group_entity is None:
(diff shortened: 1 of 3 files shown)


expertise/service/utils.py
```diff
@@ -692,31 +692,53 @@ def set_client(self, client):
         self.client = client
 
     def _generate_vertex_prefix(api_request):
```
carlosmondra, line 694: This function is doing two things:
- Validating the api request
- Returning the vertex prefix

I think the validation of the api request should be done way before we get to this point and in a separate function.
That way when we get to this function you can simply create the vertex prefix from the request without worrying about validation

haroldrubio: That's true, it makes the purpose of the function confusing, the validation does happen earlier before getting to the `create_job` function in the worker thread:
https://github.com/openreview/openreview-expertise/blob/ddcd5fb0dd2a2a05216b4af83847945ed893e1cf/expertise/service/expertise.py#L902
and the request gets validated through these two functions:
https://github.com/openreview/openreview-expertise/blob/ddcd5fb0dd2a2a05216b4af83847945ed893e1cf/expertise/service/expertise.py#L349-L359

carlosmondra (approved):

PR #234, opened 2025-04-02
Feature: Propagate Errors to API

### LINKED ISSUE #110: Expertise validation: check the get_notes by paper_invitation return a non empty result
Currently the error message is :

```
Dimension out of range (expected to be in range of [-1, 0], but got 1)
```

we should check the notes length and return an error message if the result if empty. 

This validation must be done when calling the expertise API and the command line.

### LINKED ISSUE #209: Expertise Cloud Service: Propagate Exceptions
Exceptions raised in the expertise container don't get propagated to the server. Either we can find a way to read the logs stored per job, or we can write the exception to the bucket when its caught in the container.

haroldrubio:
Resolves #209. Resolves #110.

This PR adds in new errors in `create_dataset` to detect when there is incomplete data. This avoids forcing the model to compute with incomplete data and throwing a less user friendly error.

Wraps the main computation of `execute_pipeline`, which is run in Vertex AI, in a try-catch block and writes a caught exception to the bucket.

When fetching the status of this job by ID, the error message is read from the bucket


expertise/execute_pipeline.py
```diff
@@ -28,64 +28,77 @@
...
+        execute_create_dataset(client_v1, client_v2, config.to_json())
+        execute_expertise(config.to_json())
+    except Exception as e:
+        # Write error to single JSONL line in GCS
+        error_message = {
+            'error': str(e)
+        }
+        destination_blob = f"{blob_prefix}/error.json"
+        blob = bucket.blob(destination_blob)
+        blob.upload_from_string(json.dumps(error_message))
```
carlosmondra, line 101: Wouldn't this make the job in the queue silently fail?

haroldrubio: I raised the exception again, the exception will make the pipeline return failure in Vertex, and the queue reads the error status in Vertex and makes the queue fail

expertise/execute_pipeline.py
```diff
@@ -28,64 +28,80 @@
...
+            server_config = server_config,
+            working_dir = working_dir
+        )
+
+        if working_dir is not None:
+            path_fields = ['work_dir', 'scores_path', 'publications_path', 'submissions_path']
+            config.job_dir = working_dir
+            config.dataset['directory'] = working_dir
+            for field in path_fields:
+                config.model_params[field] = working_dir
```
carlosmondra, line 88: I couldn't find these dirs in the bucket. Where are they located?
Also, we may not need to check the `if working_dir is not None:`. I think it's always true, but maybe I'm missing something.

expertise/service/expertise.py
```diff
@@ -853,35 +853,48 @@ async def worker_process(self, job, token):
...
-                    job_completed = False
-                    job_error = True
-                    break
-
-                self.logger.info(f"Job {redis_id} status: {status['status']}. Retrying in {self.poll_interval} seconds...")
+                # Check status validity
+                self.logger.info(f"INFO: before status check")
+                if status and isinstance(status, dict) and 'status' in status and 'description' in status:
+                    self.logger.info(f"INFO: after status check")
+                    config = self.redis.load_job(redis_id, user_id)
```
carlosmondra, line 865: we don't need to do this in this PR, but I suggest we get this information from the job of the queue

expertise/execute_pipeline.py
haroldrubio: Oh these paths are local to the container that get's spun up - but yeah working_dir should always be populated from an if statement before it

carlosmondra (approved):

melisabok (approved):

PR #240, opened 2025-05-07
Feature: Add support to select score by percentile

### LINKED ISSUE #232: Score calibration: Allow the parametrization of the percentile value
Let the venue organizers decide how to get the score between a submission and the reviewer's publications. 

From Nihar: 

> Instead max pooling (i.e., taking max of the submitted paper's similarity with all reviewer's papers), take a certain percentile. Max pooling will be a special case where percentile = 100.  Median will also be a special case where percentile = 50.

Add a new parameter to the request body and use 100 as a default value.

haroldrubio:
Resolves #232 

This PR adds a new config and API parameter that allows the venue to select which part of a reviewer's publication history should be used based on affinity score.

The new parameter in the API request is at the `model.percentile_select` level, or the `model_params.percentile_select` level.

A value of 100 will use the score of the most similar publication as the affinity score and a value of 0 will use the least similar publication as the affinity score for a user.

### DIFF
--- a/expertise/models/specter2_scincl/scincl.py
+++ b/expertise/models/specter2_scincl/scincl.py
@@ -40,7 +40,7 @@ silent
 """
 class SciNCLPredictor(Predictor):
     def __init__(self, specter_dir, work_dir, average_score=False, max_score=True, batch_size=16, use_cuda=True,
-                 sparse_value=None, use_redis=False, dump_p2p=False, compute_paper_paper=False):
+                 sparse_value=None, use_redis=False, dump_p2p=False, compute_paper_paper=False, percentile_select=100):
         self.model_name = 'scincl'
         self.specter_dir = specter_dir
         self.model_archive_file = os.path.join(specter_dir, "model.tar.gz")
@@ -63,6 +63,7 @@ class SciNCLPredictor(Predictor):
         self.redis = None
         self.dump_p2p = dump_p2p
         self.compute_paper_paper = compute_paper_paper
+        self.percentile_select = percentile_select
 
         self.tokenizer = AutoTokenizer.from_pretrained('malteos/scincl')
         #load base model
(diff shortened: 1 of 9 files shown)


expertise/models/specter2_scincl/scincl.py
```diff
@@ -264,7 +265,15 @@ def load_emb_file(emb_file):
...
-                if self.average_score:
+                if self.percentile_select is not None:
+                    # Select score based on percentile
+                    # q=1.0 (percentile_select=100) -> max score
+                    # q=0.5 (percentile_select=50) -> median score
+                    # q=0.0 (percentile_select=0) -> min score
+                    q = self.percentile_select / 100.0
+                    q = max(0.0, min(1.0, q)) 
+                    all_paper_aff = torch.quantile(train_paper_aff_j, q, dim=1, interpolation='linear')
+                elif self.average_score:
```
melisabok, line 276: if we can cover average_score and max_score with percentile_select, do we need to keep them? can we deprecate them?

What about calling it "percentile_score" or "percentile_select_score"?

carlosmondra: To make sure I understand.
If q=1.0 then, we only take the highest score.
if q=0.5 then, we grab the top 50% scores and average them.
if q=0.0 then, we average all scores.

Is this correct?

carlosmondra: I asked the chat and it said that it just grabs one score, there is no averaging as I had understood.

> if we can cover average_score and max_score with percentile_select, do we need to keep them? can we deprecate them?

If I understood correctly. Average score wouldn't be replaced by percentiles. @haroldrubio can you confirm 😬 ?

haroldrubio: Yes there's no averaging - I clarified with Nihar that this is the intended behavior of this feature. So we should keep averaging, but max can be achieved by setting the percentile to 1

carlosmondra: So if a reviewer has scores: 1, 0.8, 0.4, 0.1 for a particular paper and q=0.75, then, the score that would be used is 0.8. Is this correct?

haroldrubio: Yes that's correct

carlosmondra (approved):

melisabok (approved):

Issue #233, opened 2025-03-25
User's expertise: use a weight value for publications submitted to OpenReview vs imported publications (DBLP, Archive)


melisabok:
From Nihar:

> For any reviewer, when computing affinity, weigh papers authored from the same OpenReview account differently from others. The current affinity computation will be a special case where weights are equal.

One way to support this is to define three different weights, with the option of adding more in the future:

```
weights: {
 'venues': 1,
 'DBLP.org/-/Record': 0.6,
 'OpenReview.net/-/Archive': 0.4
}
```

- venues are user's publications where the invitation id prefix is member of our venue group.
- The other two have a weight based on the publication invitation id

If the publication has an invitation that is not present in the weights definition then a weight=1 is used.

@nihar

Should we specify different weights for openreview submissions that are accepted vs not accepted?

PR #241, opened 2025-05-09
Feature: Up/Downweight Publications by Venue

haroldrubio:
This PR adds a feature in the expertise system where it accepts a new field in the `dataset` field of both the API request and the JSON config called `weight_specification`. This is a list of JSON objects that specify what weight to give to which publications depending on the accepted venue.

Example for JSON config:
```
"weight_specification": [
    {
        "prefix": "NeurIPS.cc",
        "weight": 2
    },
    {
        "value": "ICLR.cc/2025/Conference",
        "weight": 2
    },
    {
        "articleSubmittedToOpenReview": true,
        "weight": 2
    }
]
```

Non-accepted papers or papers from other venues have a default weight of 1. This PR re-distributes a reviewer's affinity mass in their publication history so that more weight is given to the venues specified.

In the example above, if you provide a JSON object in the list with the key `articleSubmittedToOpenReview`, then any accepted paper from an OpenReview venue will get that weight if there isn't a match from the other objects.

Fixed https://github.com/openreview/openreview-expertise/issues/233


expertise/create_dataset.py
```diff
@@ -85,17 +87,54 @@ def deduplicate_publications(self, publications):
...
+            for venue_spec in weight_specification:
+                if not isinstance(venue_spec, dict):
+                    raise ValueError('Objects in weight_specification must be dictionaries')
+                if 'prefix' in venue_spec and 'value' in venue_spec:
+                    raise KeyError('Objects in weight_specification must only have one of [prefix, value]')
+                if 'prefix' not in venue_spec and 'value' not in venue_spec:
+                    raise KeyError('Objects in weight_specification must have a prefix or value key')
+                if 'weight' not in venue_spec:
+                    raise KeyError('Objects in weight_specification must have a weight key')
+                if not isinstance(venue_spec['weight'], int):
```
melisabok, line 134: please change this to float

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
+                    raise ValueError('Objects in weight_specification must be dictionaries')
+                if 'prefix' in venue_spec and 'value' in venue_spec and 'inOpenReview' in venue_spec:
+                    raise KeyError('Objects in weight_specification must only have one of [prefix, value, inOpenReview]')
+                if 'prefix' not in venue_spec and 'value' not in venue_spec and 'inOpenReview' not in venue_spec:
+                    raise KeyError('Objects in weight_specification must have a prefix, value, or inOpenReview key')
+                if 'weight' not in venue_spec:
+                    raise KeyError('Objects in weight_specification must have a weight key')
+                # weight must be an integer or float
+                if not isinstance(venue_spec['weight'], int) and not isinstance(venue_spec['weight'], float):
+                    raise ValueError('weight must be an integer')
```
carlosmondra, line 142: I think the following following should be:
`raise ValueError('weight must be a number')`
If we are only allowing positive numbers:
`raise ValueError('weight must be a number greater than 0')`

expertise/models/specter2_scincl/scincl.py
```diff
@@ -118,6 +119,7 @@ def set_archives_dataset(self, archives_dataset):
                     self.pub_note_id_to_abstract[publication['id']] = publication['content'].get('abstract').strip() if publication['content'].get('abstract').strip() else "."
                     pub_mdate = publication.get('mdate', int(time.time()))
                     pub_cache_key = publication['id'] + "_" + str(pub_mdate)
+                    pub_weight = publication.get('content', {}).get('weight', 1) ## Mention that default weights are 1
```
carlosmondra, line 122: I think all publications should have pub_weight at this point. The get_pub_weight method should add 1 by default so that we don't add it all the way here. It's clearer when all the logic related to how weights are gotten is in one place.

expertise/models/specter2_scincl/scincl.py
```diff
@@ -190,10 +194,11 @@ def embed_publications(self, publications_path=None):
             f.writelines(pub_jsonl)
 
     def all_scores(self, publications_path=None, submissions_path=None, scores_path=None, p2p_path=None):
-        def load_emb_file(emb_file):
+        def load_emb_file(emb_file, load_weight=False):
```
carlosmondra, line 197: Why is `load_weight` here False by default? What does it mean? Why isn't it always True, for example?

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
+        ## Papers allowed: accepted papers from an OpenReview venue
+            
+        # Find matching weight specification
+        fallback_weight = None
+        for venue_spec in weight_specification:
+            if 'prefix' in venue_spec and venueid.startswith(venue_spec['prefix']):
+                return venue_spec['weight']
+            if 'value' in venue_spec and venueid == venue_spec['value']:
+                return venue_spec['weight']
+            if 'inOpenReview' in venue_spec:
```
melisabok, line 112: what does it mean "inOpenReview"?

Is there a spec defined somewhere about the weight_specification schema?

melisabok:
It looks like the weights are only supported for SPECTER+SinCL so we should throw an error if it is being passed for another model.

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
                 deduplicated.append(pub)
 
         return deduplicated
+    
+    def get_pub_weight(self, pub=None, venueid=None, weight_specification=None):
```
melisabok, line 91: no need to pass a venueid as param, you can extract ir from the pub note, right?

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
+        if not venueid or not weight_specification:
```
melisabok, line 92: weight_specification can be empty {} as default so you do not this to do this check.

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
+            return None
+            
+        # Get domain from either domain field or invitation prefix
+        domain = getattr(pub, 'domain', None)
+        if domain is None:
+            domain = pub.invitation.split('/-/')[0]  # API1 fallback to invitation
+        
+        # Return early on DBLP papers (venueid =/= domain) and non-accepted papers (domain not in venue_list)
+        if not (venueid == domain and domain in self.venue_list):
+            return None
```
melisabok, line 102: why do you return None here? it looks like this is the inOpenReview: false rule?

expertise/create_dataset.py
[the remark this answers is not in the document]
melisabok: this probably needs to go in the expertise request so the user gets an error as soon as the request is being done, right?

melisabok: I do not see these changes in the code, why is this conversation resolved?

expertise/models/specter2_scincl/scincl.py
haroldrubio: The model shouldn't load a weight matrix if there is no `weight` field in the JSON objects of the embedding file. There are no weights for submissions and there are no weights for publications if there is no `weight_specification` field:
`load_weight=self.venue_specific_weights`

melisabok: if there is no weight_specification, aren't all the submissions and publications having weight=1? or having a weight matrix with all 1 is too expensive?

haroldrubio: If there's no `weight_specification`, there are no `weight` keys and no extra computation is done at all to preserve the original workflow

expertise/service/utils.py
```diff
@@ -475,9 +476,10 @@ def _populate_note_fields(entity, config, paper_paper_scoring=False):
         # Load optional dataset params from default config
         allowed_dataset_params = [
             'minimumPubDate',
-            'topRecentPubs'
+            'topRecentPubs',
```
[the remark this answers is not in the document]
haroldrubio: It's an older setting that says something like "keep only the 5 most recent publications"

expertise/create_dataset.py
haroldrubio: Added in recent commit, but empty [] since the schema is a list of JSON

expertise/create_dataset.py
haroldrubio: Changed in recent commit

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
         return deduplicated
+    
+    def get_pub_weight(self, pub=None, venueid=None, weight_specification=None):
+        if not venueid or not weight_specification:
+            return None
+            
+        # Get domain from either domain field or invitation prefix
+        domain = getattr(pub, 'domain', None)
+        if domain is None:
+            domain = pub.invitation.split('/-/')[0]  # API1 fallback to invitation
```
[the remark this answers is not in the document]
haroldrubio: Actually I had to revert your suggestion because it was failing, it was computing the default expression `pub.invitation.split('/-/')[0]` before trying to see if `domain` was an attribute of pub

expertise/create_dataset.py
haroldrubio: I've updated the logic, it doesn't return None now and papers following `not (venueid == domain and domain in self.venue_list)` are now selected in the `inOpenReview: false` condition

expertise/create_dataset.py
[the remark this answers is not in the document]
melisabok: Weird, I tried this and got 'NeurIPS.cc/2022/Conference'

```
note = client_v1.get_note('F0DowhX7_x')
getattr(note, 'domain', note.invitation.split('/-/')[0])
```

what is the difference?

expertise/create_dataset.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, openreview_client, openreview_client_v2, config):
             'no_publications': []
         }
 
+        self.venue_list = openreview_client_v2.get_group('venues').members
```
carlosmondra, line 38: I would make this a set so that checks like `domain in self.venue_list` are O(1) instead of O(n)

expertise/create_dataset.py
```diff
@@ -85,13 +89,50 @@ def deduplicate_publications(self, publications):
...
+
+        # Get domain from either domain field or invitation prefix
+        domain = getattr(pub, 'domain', None)
+        if domain is None:
+            domain = pub.invitation.split('/-/')[0]
+            
+        # Find matching weight specification
+        current_weight, current_order = 1, None ## Default weight one
+        for idx, venue_spec in enumerate(weight_specification):
+            if _matches(venue_spec, venueid, domain):
```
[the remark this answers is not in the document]
carlosmondra: If you do accept the suggestion, make sure to update the function

expertise/create_dataset.py
```diff
@@ -85,13 +89,50 @@ def deduplicate_publications(self, publications):
...
+                order = venue_spec.get('order', idx) ## Support optional order, fallback to index
+                if current_order is None or order <= current_order:
+                    current_weight = venue_spec['weight']
+                    current_order = order
```
carlosmondra, line 126: What is `order` used for?

expertise/service/utils.py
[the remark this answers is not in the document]
enrubio: Okay, I included this in my PR.

expertise/create_dataset.py
carlosmondra: I saw above your comment :) 
If you can share an example, it would be helpful

expertise/create_dataset.py
```diff
@@ -85,13 +89,50 @@ def deduplicate_publications(self, publications):
...
+                ('inOpenReview' in venue_spec and not in_openreview and not venue_spec['inOpenReview'])
+            )
+        
+        venueid = pub.content.get('venueid', '')
+        if isinstance(venueid, dict):
+            venueid = venueid.get('value')
+        
+        ## If weight specification but no venueid, return default weight
+        if not venueid:
+            return 1
```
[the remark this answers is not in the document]
haroldrubio: Melisa made this comment:
> weight_specification can be empty {} as default so you do not this to do this check.

I understood this as let the function return a value but don't set the venue weight. I previously added the check to avoid running this function, I could also check `weight_specification` here to avoid calling the function entirely:
https://github.com/openreview/openreview-expertise/blob/358158a3edbd382c622c43c3cd1ebda877d46552/expertise/create_dataset.py#L168

expertise/create_dataset.py
haroldrubio: Implementing value > prefix > articleSubmittedToOpenReview rule

expertise/create_dataset.py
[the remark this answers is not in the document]
haroldrubio:
```
note = client.get_note('iGcxlTLIL5')
getattr(note, 'domain', note.invitation.split('/-/')[0])
```
It fails for me on API2 notes

expertise/create_dataset.py
haroldrubio: Logic defined for new `articleSubmittedToOpenReview` key in discussion

expertise/create_dataset.py
```diff
@@ -85,17 +87,59 @@ def deduplicate_publications(self, publications):
...
+            
+        # Find matching weight specification
+        fallback_weight = None
+        for venue_spec in weight_specification:
+            if 'prefix' in venue_spec and venueid.startswith(venue_spec['prefix']):
+                return venue_spec['weight']
+            if 'value' in venue_spec and venueid == venue_spec['value']:
+                return venue_spec['weight']
+            if 'inOpenReview' in venue_spec:
+                fallback_weight = venue_spec['weight']
```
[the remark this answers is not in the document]
haroldrubio: Discussion resolved in later comment by implementing value > prefix > articleSubmittedToOpenReview

expertise/create_dataset.py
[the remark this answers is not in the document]
haroldrubio:
```
Cell In[3], line 2
      1 note = client.get_note('iGcxlTLIL5')
----> 2 getattr(note, 'domain', note.invitation.split('/-/')[0])

AttributeError: 'Note' object has no attribute 'invitation'
```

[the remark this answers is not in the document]
melisabok: I think what is happening is that the default value param code is executed before the getattr function. Because API 2 notes don't have the attribute invitation then it fails. One workaround would be:

`getattr(note, 'domain', getattr(note, 'invitation', note.invitations[0]).split('/-/')[0])`

or just leave the if sentence.

PR #249, opened 2025-07-07
Testing: Use GCS bucket

haroldrubio:
This PR switches from mocking GCS for integration tests to using a test folder in a GCS bucket. The PR keeps mocking the bucket in the GCP interface tests since these are unit tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ orbs:
 jobs:
   build:
     resource_class: large
+    parallelism: 10
     working_directory: ~/openreview-expertise
     docker:
       - image: ubuntu:22.04


tests/conftest.py
```diff
@@ -328,6 +332,50 @@ def post_editor_data(client, data, editors):
...
+    @staticmethod
+    def cleanup_job_files(bucket):
+        """Clean up job files from the bucket"""
+        if bucket is None:
+            return
+        
+        # Handle both real GCS buckets and mock buckets
+        blobs = list(bucket.list_blobs(prefix=f"{GCSTestHelper.GCS_JOBS_FOLDER}/"))
+        for blob in blobs:
+            blob.delete()
```
carlosmondra, line 361: Do they need to be deleted?

haroldrubio: Should we keep old test data in the bucket and just control the files with a lifecycle?

carlosmondra: Yeah, we can do that.

.circleci/config.yml
```diff
@@ -12,7 +12,6 @@ jobs:
   build:
     # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
     resource_class: large
-    parallelism: 10
     working_directory: ~/openreview-expertise
```
carlosmondra, line 15: I remember you explained why we had to remove the parallelism, but can you remind me? That way it will be documented too. Thanks!

carlosmondra (approved):

PR #250, opened 2025-07-10
Cloud: More efficient resource allocation

haroldrubio:
This PR computes `notes_count` in the worker thread based on the config (initially: using only submissions or alternate group publications) and adds this to the request that gets sent to the pipeline. The pipeline reads the field to determine which resources should be used.

[Documentation](https://www.kubeflow.org/docs/components/pipelines/user-guides/core-functions/control-flow/) suggests using control flow from the library because the `@pipeline` decorated code is an orchestration layer rather than true code - regular if/elses in that function wouldn't appear when it get compiled into the pipeline definition

Note: Take into account that a machine type might be too small, we might need a way to fallback to a larger machine size to make sure that jobs won't continuously fail by lack of resources


expertise/build_pipeline.py
```diff
@@ -84,6 +86,21 @@ def write_string_to_file_op(
...
+        config = json.loads(job_config)
+
+        paper_count = len(config.get('notes_count', 0))  # Adjust based on your actual config
+
+        ## TODO: Maybe make a file to store this configuration?
+        ## Same with resource specifications
+        if paper_count < 1000:
+            return "small"
+        else:
+            return "large"
```
carlosmondra, line 102: We don't have config file for the expertise API?

expertise/build_pipeline.py
```diff
@@ -93,9 +110,19 @@ def execute_expertise_pipeline_op(
...
-    custom_expertise_job_from_file_input = create_custom_training_job_from_component(
+    small_expertise_job_from_file_input = create_custom_training_job_from_component(
         execute_expertise_pipeline_op,
-        display_name="expertise-job",
+        display_name="expertise-job-small",
+        machine_type="n1-standard-16",
+        accelerator_type="NVIDIA_TESLA_T4",
+        accelerator_count=1,
+        boot_disk_type="pd-ssd",
+        boot_disk_size_gb=200,
```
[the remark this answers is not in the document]
haroldrubio: Oh yes I can reduce it for the smaller machine type, the jobs still need some amount of disk because they write the embeddings to disk

expertise/build_pipeline.py
carlosmondra: Can you explain why the Redis/BullMQ config cannot be used for this script?

melisabok: I wonder if superuser can force the job_size.

carlosmondra: Maybe we should use a single config file for everything. Is there a reason why we should separate configuration files?

expertise/build_pipeline.py
```diff
@@ -84,6 +86,21 @@ def write_string_to_file_op(
...
+    def determine_job_size_op(job_config: str) -> str:
+        """Determines job size based on paper count"""
+        import json
+        config = json.loads(job_config)
+
+        paper_count = len(config.get('notes_count', 0))  # Adjust based on your actual config
+
+        ## TODO: Maybe make a file to store this configuration?
+        ## Same with resource specifications
+        if paper_count < 1000:
```
melisabok, line 99: I wonder if need three categories. 

< 1000
between 1k and 10k
> 10k

>10k will be used for the 6 large venues that we host so far: NeurIPS, ICML, ICLR, CVPR, AAAI and ICCV

haroldrubio:
Yeah I think this can work. The `create_dataset` job would fetch and upload the data to the temp folder in GCS. We'll save GPU time but it would add a little more overhead in terms of time per job:
1) Downloading the archives data into the GPU job
2) Extra setup/teardown of the `create_dataset` job

I think the more interesting use case is once we enable an embedding cache, the CPU job can fetch the dataset and existing embeddings and only compute missing embeddings if there are any. Then it can do the final steps of the score computation on CPU

haroldrubio:
> I don't know how long this would take, but in this PR we are downloading the notes already to compute the size of the instance and then we do it again in the create_dataset. I assume that takes more time than the steps above and it also means less stress for the OpenReview API. So I would like to optimize it. If it's too much work, for now, we can merge this PR first and then do the improvements. What do you think?

This makes sense but maybe it should go in a separate PR since it would be possible to separate out this step without adding in the extra feature of selectable machine types. But also I'm liking this idea more because if we can do this, then we should be able to get away with not needing the token in the job data.

> Another option is to run create_dataset in the VM instance just before creating the Vertex AI pipeline. Would this be easier?

I was thinking about this in regards to what I said previously - my biggest thing is that for NeurIPS-sized conferences, doing this synchronously would take on the order of minutes to fetch. And doing this asynchronously with another kind of job in the queue

carlosmondra:
> Yeah we could do something like that:
> 
> 1. Check that the files aren't already in GCS
> 2. Retrieve the dataset and load it onto disk
> 3. Upload the dataset to GCS (maybe there's a new folder in the `openreview-expertise` bucket called `dataset`)
> 4. Delete the files from disk
> 5. Continue with `create_job`
> 
> Is this close to what you had in mind?

Yes, how much space do the files occupy in the instance? We need to make sure we have enough room. Then, the create_job could check the amount of data we have to calculate the size of the instance needed.

> Unfortunately, this still doesn't avoid sending the token in the job data but maybe we can think about resolving that issue separately

Yes, but I wouldn't try to fix that in this PR.

carlosmondra (approved):

expertise/service/config/default.cfg
```diff
@@ -29,3 +29,28 @@ DEFAULT_CONFIG = {
         "use_redis": False
     }
 }
+
+SMALL_NAME = 'small'
+MEDIUM_NAME = 'medium'
+LARGE_NAME = 'large'
+
+PIPELINE_MEDIUM_THRESHOLD = 1000
```
melisabok, line 37: it looks like you are only count the submissions of a venues to decide the machine type, correct if I'm wrong:

1. does it mean that TMLR that computes scores per submission is going to always use the SMALL pipeline?
2. if we compute affinity scores of ACs vs SACs, is it using a SMALL pipeline also?
3. looking at the openreview.net/venues.html stats, 90% of the venues have less than 100 submission, should we use this number for the small threshold?
4. do you have some stats that show the distribution of machine types for all our existing venues? I would like to know if the frequency of each machine type.
5. are retrying the job with a higher machine type if it runs out of memory?
6. One suggestion about the machine type configuration: use a dictionary to the define the different machine type and thresholds and pipelines so if we want to add a fourth machine type then we can only change the configuration.

melisabok (approved):

PR #252, opened 2025-07-15
Chore: Update Github Action to new instance

haroldrubio:
This PR points the action to the new instance and runs a separate script to update openreview-py

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -33,7 +33,8 @@ jobs:
           export_environment_variables: true
       - name: Setup gcloud
         uses: google-github-actions/setup-gcloud@v1
-      - name: Run deploy script
+      - name: Run deploy scripts
         run: |
-          gcloud config set compute/zone us-central1-c
-          gcloud compute ssh openreview@instance-matching-server-gpu --command '/bin/expertise_build.sh' --quiet
+          gcloud config set compute/zone us-central1-b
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -33,7 +33,8 @@ jobs:
           export_environment_variables: true
       - name: Setup gcloud
         uses: google-github-actions/setup-gcloud@v1
-      - name: Run deploy script
+      - name: Run deploy scripts
         run: |
-          gcloud config set compute/zone us-central1-c
-          gcloud compute ssh openreview@instance-matching-server-gpu --command '/bin/expertise_build.sh' --quiet
+          gcloud config set compute/zone us-central1-b
+          gcloud compute ssh openreview@openreview-expertise-instance --command '/bin/openreview-py_build.sh' --quiet
```
carlosmondra, line 39: Is it OK to always deploy openreview-py when we deploy the expertise?

melisabok: I don't think so, we also need to deploy openreview-py from the tag version not master.

melisabok: I'm not sure, Carlos, should it be here or in the openreview-py repo? https://github.com/openreview/openreview-py/actions I see that is how Celeste deploys openreview-py to the APIs

carlosmondra (approved):

PR #259, opened 2025-07-30
Feature: Allow disable normalization

haroldrubio:
This PR adds a new flag `normalize_scores` to the `model_params` field of the config file that allows the user to skip the normalization across all computed scores.

The default behavior is to normalize scores in order preserve the current functionality


carlosmondra (approved):
Thanks for putting this together so quickly. It looks good to me!

PR #289, opened 2025-09-18
Fix: Prevent "Initialized" config in Redis

### LINKED ISSUE #287: Vertex AI: Requesting expertise creates second job that doesn't run
A user reported that requesting the expertise creates a second imaginary job with the status of 'initialized' which would would never run.

haroldrubio:
Resolves #287

This PR resolves an issue where `compute_machine_type` adds another Redis config because the validation function in `expertise/service/expertise.py` to transform an `APIRequest` into a `JobConfig` is where the Redis config is first stored.


expertise/service/expertise.py
```diff
@@ -368,13 +371,14 @@ def _prepare_config(self, request, job_id=None, client_v1=None, client=None) ->
...
-            self.logger.info(f"Saving processed config to {os.path.join(config.job_dir, 'config.json')}")
-            self.redis.save_job(config)
+        if save:
+            if not os.path.isdir(config.dataset['directory']):
+                os.makedirs(config.dataset['directory'])
+            with open(os.path.join(config.job_dir, 'config.json'), 'w+') as f:
+                json.dump(config.to_json(), f, ensure_ascii=False, indent=4)
+            if not self.containerized:
+                self.logger.info(f"Saving processed config to {os.path.join(config.job_dir, 'config.json')}")
+                self.redis.save_job(config)
```
melisabok, line 381: why don't we save the job by id in redis so we can override it if has the same id?

haroldrubio: They do get saved by `config.job_id` but if `JobConfig.from_request` gets a None `job_id` or doesn't get passed a `job_id`, it creates generates a new one

melisabok: why the job_id is None?

haroldrubio: The job_id is None when we need to make a new configuration for a new job, do you mean for here when I need to parse the request again so that I can get the number of submissions?
https://github.com/openreview/openreview-expertise/blob/547a7dfc8892751ba99e7378485c7dda66934d4a/expertise/service/expertise.py#L833-L838
Yeah actually I think I can just pass the job_id here

expertise/service/expertise.py
```diff
@@ -336,6 +336,9 @@ def _prepare_config(self, request, job_id=None, client_v1=None, client=None) ->
         :param job_id: If provided, use this job ID instead of generating a new one
         :type job_id: str
 
+        :param save: If True, save the job config to Redis and disk after preparing it
+        :type save: bool
+
```
carlosmondra, line 341: I don't see this param being passed anymore, so I suppose it can be removed

expertise/service/config/default.cfg
```diff
@@ -10,6 +10,7 @@ REDIS_ADDR = 'localhost'
 REDIS_PORT = 6379
 REDIS_CONFIG_DB = 10
 REDIS_EMBEDDINGS_DB = 11
+SYNC_ON_DISK = False
```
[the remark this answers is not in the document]
melisabok: why do you need to use a different value and in which scenarios?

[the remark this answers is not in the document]
haroldrubio: This was from when we were running the service on a single server with a GPU - we had a cronjob that deleted the job files to save space but the cronjob didn't delete it from Redis. So, to prevent the deleted jobs from being returned by the API, we skipped jobs that didn't have a job file on disk.

expertise/service/expertise.py
```diff
@@ -825,15 +821,19 @@ def set_client_v2(self, client_v2):
         self.client_v2 = client_v2
         self.cloud.set_client(client_v2)
 
-    def compute_machine_type(self, client, client_v2, api_request):
-        config, _ = self._prepare_config(deepcopy(api_request), client_v1=client, client=client_v2)
+    def compute_machine_type(self, client_v1, client, job_id):
+        config, _ = self._validate_request(
```
melisabok, line 825: You don't need to call this function to get the config, you can call to `self.redis.load_job(job_id, get_user_id(client))` directly.

expertise/service/expertise.py
```diff
@@ -366,17 +360,18 @@ def _prepare_config(self, request, job_id=None, client_v1=None, client=None) ->
             working_dir = self.working_dir
         )
         self.logger.info(f"Config validation passed - {config.to_json()}")
+        return config, or_client.token
```
melisabok, line 363: why do you need to run the token here? can you get the token after calling this function?

expertise/service/expertise.py
```diff
@@ -327,30 +327,16 @@ def _filter_config(self, running_config):
         running_config.baseurl_v2 = None
         running_config.user_id = None
 
-    def _prepare_config(self, request, job_id=None, client_v1=None, client=None) -> dict:
+    def _validate_request(self, client_v1, client, request, job_id=None):
```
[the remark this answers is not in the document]
haroldrubio: Fixed in recent commit

expertise/service/expertise.py
```diff
@@ -639,7 +628,6 @@ def start_expertise(self, request, client_v1, client):
         self.logger.info('just before submitting')
 
         self.logger.info(f"\nconf: {config.to_json()}\n")
-        self.redis.save_job(config)
```
melisabok, line 642: you removed the save here but you are changing the status to QUEUED? and then it is set later in another call, when is the job really queued?

melisabok (approved):

carlosmondra (approved):

PR #298, opened 2025-10-03
Feature: Loading vector cache secret

haroldrubio:
This PR to the embeddings cache PR integrates a Google Secrets Manager call into the pipeline. The environment variable and the secret name are passed as sibling fields to the `gcs_request_path` so that they can be accessed at runtime.


expertise/build_pipeline.py
```diff
@@ -133,11 +133,33 @@ def _coerce_numeric(value):
...
+        mongodb_collection: str,
+        secret_version: str = "latest",
+        project_id: str = ""
     ) -> None:
+        import os
+        from google.cloud import secretmanager
+
+        # Fetch MongoDB URI secret at runtime
+        if not project_id:
+            raise RuntimeError("project_id must be provided to access Secret Manager")
```
carlosmondra, line 152: Does this mean that this change is not backward compatible?

expertise/service/config/default.cfg
```diff
@@ -13,6 +13,8 @@ REDIS_EMBEDDINGS_DB = 11
 MONGODB_URI = 'mongodb://localhost:27017'
 MONGO_EMBEDDINGS_DB = 'openreview_test'
 MONGO_EMBEDDINGS_COLLECTION = 'openreview_embeddings'
+MONGODB_SECRET_ID = 'mongodb-uri'
```
carlosmondra, line 16: This is the name of the secret right?

expertise/build_pipeline.py
carlosmondra: I guess I don't understand why you raise an exception if project_id is not present. I think we should just skip it.

Issue #302, opened 2025-10-20
Sparse scores not returned correctly


enrubio:
This was tested using paper similarity scores. 

It seems like the sparse value is either not being read from the expertise request or it's not being returned correctly by `get_expertise_results`. It looks like the full scores are being returned.

Example job IDs for a venue with `1867` submissions: 
- `xzgzzsfkgj` - sparse value = 5
- `o9efv1q3p2` - sparse value = 20

Both return `3,485,689` scores, which is exactly `1867 x 1867`.

melisabok:
@enrubio can you run these jobs again but using a different model and see if you still get the same number of results?

melisabok:
Thanks for checking, now if try this:

1. run paper-paper scores using two different venues and a sparse value
2. run paper-paper scores using the same venue and a sparse value

What do you get in each scenario? all the scores or the requested sparse value?

melisabok:
Thanks for checking, Carlos will take a look.

Issue #277, opened 2025-09-10
SAC-AC affinity scores - sparse file


celestemartinez:
When retrieving scores between SACs and ACs for ICLR 2026, I noticed that the sparse file does not contain SAC-AC scores, but scores between papers and users. For example:

> {"match_member": "zzyGkoeDAg", "submission_member": "~James_Kwok1", "score": 0.9074164032936096}
{"match_member": "zzyGkoeDAg", "submission_member": "~Sung_Ju_Hwang1", "score": 0.8217486441135406}
{"match_member": "zzyGkoeDAg", "submission_member": "~Jose_Miguel_Hernandez_Lobato1", "score": 0.8175722062587738}
{"match_member": "zzyGkoeDAg", "submission_member": "~Michael_M._Bronstein1", "score": 0.7944488823413849}

melisabok:
I do think we need to compute sparse scores for group vs group the same way we do it for notes vs groups. There can be large groups and we don't want to upload all the edges.

Can you use another temp file to keep the full scores?

PR #305, opened 2025-10-28
Fix/sparse scores

carlosmondra:
This pull request refactors how sparse score computation is handled in the expertise prediction pipeline. The logic for generating sparse scores, which was previously duplicated across several model classes, is now centralized into a single utility function. This reduces code duplication and simplifies the main execution flow. Additionally, the main execution function is updated to use a consistent predictor variable and to handle sparse score generation in a unified way for all models.

Refactoring and code simplification:

* Centralized sparse score generation by removing the `sparse_scores` methods (and their helpers) from individual model classes (`bm25.py`, `ensemble.py`, `multifacet_recommender.py`) and replacing them with a new `generate_sparse_scores` utility function, now imported and used in `execute_expertise.py`. [[1]](diffhunk://#diff-8fc93d87e492c9adb9e924cea794fc03b63cc8d388532131c2f76b90b719c3a6L131-L171) [[2]](diffhunk://#diff-3333923b3faae80f76fa598862239f7e512b6730429be22a5bf463abd9fcd0baL94-L138) [[3]](diffhunk://#diff-980a917b4bf6d6e59d4e94010760011ea180991df694c5a449293875c1355a26L924-L968) [[4]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L6-R6)
* Updated the `execute_expertise` function to use a single `predictor` variable for all models, reducing repeated code and improving readability. [[1]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L22-R41) [[2]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L57-R68) [[3]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L91-R103) [[4]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L131-R134) [[5]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L167-R165) [[6]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L199-R186) [[7]](diffhunk://#diff-5afe892fd76b6f994a639a37b3012f3213fe2fbf4b00031cce0da991f366e602L231-R229)

Sparse score handling improvements:

* Unified the logic for generating sparse scores after all models' main score computation, using the new utility function and handling the `sparse_value` parameter in a single place.
* Ensured that when an alternate match group is specified in the config, the correct preliminary scores are used for sparse score generation.

These changes make the codebase easier to maintain and extend, and ensure consistent sparse score computation across all expertise models.

Fixes https://github.com/openreview/openreview-expertise/issues/302
Fixes https://github.com/openreview/openreview-expertise/issues/277

### DIFF
--- a/tests/test_expertise_apiv2.py
+++ b/tests/test_expertise_apiv2.py
@@ -856,7 +856,6 @@ class TestExpertiseV2():
         assert response['description'] == 'Job is complete and the computed scores are ready'
 
         results = test_client.get('/expertise/results', headers=openreview_client.headers, query_string={'jobId': job_id}).json['results']
-        print('Results:', results)
         assert len(results) == 25 # 5 submissions x 5 submissions/publications from Raia/Kyunghyun
 
         sorted_results = sorted(results, key=lambda x: x['score'], reverse=True)
@@ -915,7 +914,6 @@ class TestExpertiseV2():
         assert response['description'] == 'Job is complete and the computed scores are ready'
 
         sparse_results = test_client.get('/expertise/results', headers=openreview_client.headers, query_string={'jobId': job_id}).json['results']
-        print('Sparse Results:', sparse_results)
         assert len(sparse_results) == total_submissions # Sparse value of 1
 
         sorted_sparse_results = sorted(sparse_results, key=lambda x: x['score'], reverse=True)


expertise/execute_expertise.py
```diff
@@ -228,33 +198,35 @@ def execute_expertise(config):
...
-        if config['model_params'].get('sparse_value'):
-            ens_predictor.sparse_scores(
-                scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '_sparse.csv')
-            )
-
     if 'alternate_match_group' in config.keys():
-        aggregate_by_group(config)
+        preliminary_scores = aggregate_by_group(config)
+    else:
+        preliminary_scores = predictor.preliminary_scores
```
melisabok: this else is for scores where "notes vs groups" and "notes vs notes"?

the only place we need to group the sores is for "groups vs groups". Am I right?

carlosmondra: That's my understanding. @haroldrubio can you confirm?

expertise/service/utils.py
```diff
@@ -1219,71 +1214,49 @@ def _get_scores_and_metadata_streaming(all_blobs, job_id, group_scoring=False, p
...
+
                 # Yield any remaining results
                 if results_chunk:
                     yield {'results': results_chunk, 'metadata': None}
-                    
+
                 downloader.close()
+
+            # Determine whether to use sparse or non-sparse score files
+            sparse_score_files = [
```
melisabok: Why do we have an array of files? is there a chance we need to stream more than one file?

carlosmondra: I don't think that's ever supposed to happen. I think @haroldrubio has it as some form of validation. If there's more than one file with the `sparse` substring, it fails. I am not sure why this would happen though. @haroldrubio, is my interpretation correct?

haroldrubio: I don't think it's supposed to happen, we can filter by "_sparse" or maybe build the filename directly? I think I implemented it this way because I wasn't sure what the name of the file was going to be but I think it should be easier now

melisabok (approved):

PR #312, opened 2025-11-17
Feature: Deduplicate to paper with abstract

haroldrubio:
This PR builds a reviewer's expertise by the following priority:
```
        Priority 1: Abstract presence (non-zero abstract wins)
        Priority 2: Date (more recent wins)
        If equal: return False (keep existing, arbitrary tie-break)
```

Changes are made to expertise building for both the match group and alternate match group logic

### DIFF
--- a/tests/test_create_dataset.py
+++ b/tests/test_create_dataset.py
@@ -888,6 +888,36 @@ def test_expertise_inclusion(client, openreview_client, helpers, clean_start_con
 def test_paperhash_deduplication_priority(client, openreview_client, helpers):
     author_id = '~Harold_Rice1'
     test_title = "Test Paperhash"
+
+    # Make abstract optional in openreview.net/Archive/-/Direct_Upload invitation
+    openreview_client.post_invitation_edit(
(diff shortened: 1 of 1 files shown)


tests/test_create_dataset.py
```diff
@@ -883,4 +883,127 @@ def test_expertise_inclusion(client, openreview_client, helpers, clean_start_con
...
+        },
+        pdate=1554819115,  # Older date
+        license='CC BY-SA 4.0'
+    )
+    
+    # Paper 4: Newest date, with abstract (should be kept)
+    paper4 = openreview.api.Note(
+        content={
+            "title": {'value': test_title},
+            "abstract": {'value': 'This is the best abstract with the newest date.'},  # Has abstract
```
melisabok: can you add a test where the field abstract is not in content?

haroldrubio:
I was adding it this morning but it seems like my tests are failing in API1 on user registration, I wonder if it's because of the Cloudflare outage because it's failing to parse a POST request to the turnstile URL

PR #343, opened 2026-03-31
implementing job-splitting plan from #340

cagostino:
## Summary

Implement the job-splitting plan from #340: split the expertise BullMQ job into Task 1 (create dataset locally in the worker) and Task 2 (submit embedding + scoring to Vertex AI). Task 1 fails fast on empty data and computes machine type from the actual dataset; Task 2 runs on one of three per-tier compiled pipelines (small/medium/large) and reads the pre-built dataset straight from GCS.

### Split `worker_process` into Task 1 (dataset) + Task 2 (Vertex)

Refactor `ExpertiseCloudService.worker_process` so data fetching from the OpenReview API happens in the BullMQ worker (CPU-only, no GPU idle time) and only embedding + scoring runs on Vertex. The worker uploads the dataset to GCS, picks the tier from the on-disk metadata, then submits the per-tier pipeline.

- **`expertise/service/expertise.py`**
  - `worker_process` now runs `execute_create_dataset` first, reports `DATA_ERROR` on `ExpectedDataError` and `ERROR` on anything else, and only then uploads + submits to Vertex.
  - Replaced `compute_machine_type` (which re-queried the API from the request) with `compute_machine_type_from_dataset`, which reads `submission_count + archives_count` from `metadata.json` written by `execute_create_dataset`.
  - `cloud_id` is generated up front as `{job.id}-{ms}` so both the uploaded dataset and the Vertex pipeline land in the same GCS folder (fixes the `get_job_results` lookup-in-wrong-folder bug).
  - Dropped the v1 client plumbing from `_validate_request`, `start_expertise`, and the v2-only `execute_create_dataset` call; `baseurl_v1` / `set_client` / the v1 `Client` instantiation are all removed.
  - Persist `QUEUED` in `start_expertise` so clients see the status at enqueue time; dropped the redundant re-sets inside `worker_process`.

- **`expertise/service/utils.py`**
  - Added `GCPInterface.upload_dataset(config, vertex_id)` — uploads `archives/`, `submissions/`, `submissions.json`, and `metadata.json` under `{jobs_folder}/{vertex_id}/dataset/` using `storage.transfer_manager.upload_many_from_filenames` (parallel upload, not a per-file loop).
  - `GCPInterface.create_job` accepts `dataset_gcs_path` and `vertex_id`, passes the former into `parameter_values`, and reuses the caller's `vertex_id` instead of generating its own. Also rebuilt the `APIRequest` from explicit fields so unknown keys can't leak into the pipeline request.
  - Selects the per-tier pipeline at submit time via `pipeline_name_by_tier[machine_type]` (built from `SMALL_NAME`/`MEDIUM_NAME`/`LARGE_NAME` + base `pipeline_name`) instead of one pipeline with in-pipeline branching.
  - `get_job_results`: filter `metadata.json` and score files to exclude anything under `dataset/`, and validate file presence **before** returning the streaming generator so a missing scores folder raises `OpenReviewException` up front instead of producing a broken chunked response mid-stream.
  - Removed the now-unused `EXPERTISE_QUEUED` status (reuse `QUEUED`) and dropped the v1-fallback branches in `JobConfig.from_request` (invitation label lookup is v2-only).
  - Token is no longer written into the GCS `request.json` — the Vertex container receives it via the pipeline client instead.

### Three per-tier KFP pipelines instead of one branching pipeline

- **`expertise/build_pipeline.py`**
  - Compile and upload three pipelines per tag — `{kfp_name}-small`, `-medium`, `-large` — each wrapping the corresponding `{small,medium,large}_expertise_job_from_file_input` component. Removed the `If` / `Elif` / `Else` branching on `machine_type`.
  - Each pipeline accepts `gcs_request_path` and `dataset_gcs_path`; no `machine_type` parameter (Vertex rejects the job if an undeclared parameter is passed).
  - Upload each tier separately with its own `version_exists` guard against `RegistryClient`.

### Vertex container reads the pre-built dataset

- **`expertise/execute_pipeline.py`**
  - New `dataset_gcs_path` arg: when present, `download_dataset_from_gcs` pulls the dataset under the working dir and `execute_create_dataset` is skipped entirely — Vertex runs `execute_expertise` only.
  - `load_model_artifacts` now downloads only the artifacts the requested model needs (`artifacts_for_model(name)`), not all of them on every cold start.
  - Always dump embeddings to the bucket (`dump_embs` flag removed); archives dump is still gated on `dump_archives` and the directory existing.
  - Parses `APIRequest` from explicit fields so any unexpected keys at the top of the GCS request.json are rejected.

### Dataset-size metadata and tier thresholds

- **`expertise/create_dataset.py`**
  - Record `metadata['archives_count'] = len(expertise)` so the worker can compute the tier without re-counting files.
- **`expertise/service/config/default.cfg`**
  - Doubled the thresholds — `PIPELINE_MEDIUM_THRESHOLD = 2000`, `PIPELINE_LARGE_THRESHOLD = 10000` — because the new counter sums submissions from entity A **and** entity B (plus archives), not just one side.

### Tests

- **`tests/test_pipeline.py`**
  - New tests for `dataset_gcs_path`: download path invoked when provided, skipped when absent; metadata/archives fixtures wired up in every pipeline test that inspects the GCS output; `ExpectedDataError` path verifies `expected: true` in `error.json`.
- **`tests/test_gcp_interface.py`**
  - `upload_dataset` coverage: archives, submissions, and `submissions.json` are passed to `transfer_manager` with the right bucket prefix; empty-directory case is a no-op.
  - `create_job` assertions: per-tier pipeline URL is used based on `machine_type`; `machine_type` is **not** in `parameter_values` (regression guard for the Vertex rejection); `token` is absent from the GCS request.json.
- **`tests/test_expertise_cloud_service.py`**
  - After each job, list blobs under `dataset/` in the live bucket and assert the expected files are present.
  - Switched cloud-service group fixtures to `CLD.cc/Reviewers` (posted via `DEFAULT_POST_AREA_CHAIRS=False`) so Task 1 actually finds publications instead of tripping `Data Error`.
  - Use `tmlr_client` headers for TMLR job requests since the `CLD.cc` client lacks cross-venue read access.

## Why

- **Fail fast on empty data.** Before this PR, an empty dataset would only surface after the Vertex pipeline had spun up a GPU VM and started. Task 1 validates the dataset in the worker and returns `DATA_ERROR` immediately.
- **Don't idle a GPU on an API download.** The previous container downloaded from the OpenReview API before running the models — a T4 (and, at the large tier, 4x T4) sitting idle for the duration of the fetch. Task 2 now only runs compute.
- **Machine type from actual data.** Tier selection used to estimate size from request params; now it reads `submission_count + archives_count` out of the dataset metadata that Task 1 already writes.
- **Remove dead v1 paths.** The service-side consolidation onto v2 in #354 / #355 made the v1 fallbacks in `_validate_request`, `JobConfig.from_request`, `worker_process`, and `create_dataset` dead code. They're dropped here.

## Deployment notes

- The deploy workflow now produces **three** KFP pipeline artifacts per tag (`{base_name}-small`, `-medium`, `-large`) instead of one. The first deploy after this merges will upload all three; subsequent deploys with the same tag will no-op per tier via the existing `version_exists` guard.
- `development.cfg` / `production.cfg` do not need edits — `pipeline_name_by_tier` is derived from `SMALL_NAME` / `MEDIUM_NAME` / `LARGE_NAME` + the existing base `pipeline_name`.
- `PIPELINE_MEDIUM_THRESHOLD` / `PIPELINE_LARGE_THRESHOLD` were doubled in `default.cfg`; any env overrides on the VMs should be bumped proportionally or left unset to pick up the defaults.
- Any Redis jobs mid-flight at deploy with the old `EXPERTISE_QUEUED` status will show up with no matching description entry; drain the queue before deploying if that matters.


tests/test_expertise_cloud_service.py
melisabok: this should be a wrong conflict resolution, you need to use tmlr_client here otherwise you get an error

melisabok: same here and all the requests related to the TMLR job

expertise/service/expertise.py
```diff
@@ -864,22 +844,36 @@ async def worker_process(self, job, token):
         config.mdate = int(time.time() * 1000)
         config.status = JobStatus.QUEUED
         config.description = descriptions[JobStatus.QUEUED]
-        openreview_client_v2 = openreview.api.OpenReviewClient(
-            token=or_token,
-            baseurl=config.baseurl_v2
-        )
+        openreview_client_v2 = openreview.api.OpenReviewClient(token=or_token, baseurl=config.baseurl_v2)
```
melisabok, line 836: given that the client is created only to create the dataset, we can move this inside of the "execute_create_dataset" and pass the config as the only param. Is the token stored in the config?

expertise/service/expertise.py
```diff
@@ -336,7 +334,18 @@ def _validate_request(self, client, request):
         or_client = client if client else self.client_v2
 
         self.logger.info(f"Incoming request - {request}")
-        validated_request = APIRequest(request)
+        allowed_fields = {'name', 'entityA', 'entityB', 'model', 'dataset', 'machineType'}
+        unexpected = set(request.keys()) - allowed_fields
+        if unexpected:
+            raise openreview.OpenReviewException(f"Bad request: unexpected fields in request: {sorted(unexpected)}")
```
carlosmondra: Why are we making this change? We can keep it, but it feels like it's outside the scope of the PR and I would do this using pydantic instead. We should create an issue for this and address it separately.

expertise/service/utils.py
```diff
@@ -910,7 +895,32 @@ def _generate_vertex_prefix(api_request):
...
+        for item in dataset_items:
+            local_path = os.path.join(job_dir, item)
+            if os.path.isdir(local_path):
+                for root, dirs, files in os.walk(local_path):
+                    for fname in files:
+                        file_path = os.path.join(root, fname)
+                        relative_path = os.path.relpath(file_path, job_dir)
+                        blob = self.bucket.blob(f"{folder_path}/{relative_path}")
+                        blob.upload_from_filename(file_path)
+                        self.logger.info(f"Uploaded {relative_path} to gs://{self.bucket_name}/{folder_path}/{relative_path}")
```
carlosmondra: I am still worried about these nested for loops. Is this OK?

expertise/service/expertise.py
cagostino: I will remove it and agree that we should be using pydantic instead for this. 

as for why this was changed,  the original `APIRequest.__init__` took a raw dict and validated unknown fields implicitly by consuming known fields via `pop` and then checking what was left. In the pre-refactor code (`expertise/service/utils.py`, lines 112–113 before commit `65f5b6e`): original `APIRequest.__init__` took a raw dict and validated unknown fields
**implicitly** by consuming known fields via `pop` and then checking what was left.


In the pre-refactor code ([`expertise/service/utils.py` lines 112–113](https://github.com/openreview/openreview-expertise/blob/0af555868aa1ee29488538f338612898149b85a0/expertise/service/utils.py#L112-L113)):

```python
# Old APIRequest.__init__(self, request) — request was mutated by popping known fields
if len(request.keys()) > 0:
    raise openreview.OpenReviewException(
        f"Bad request: unexpected fields in {root_key}: {list(request.keys())}"
    )
```

Melisa asked us in slack to stop passing a raw dict and instead pass explicit params (`APIRequest(entityA, entityB, model, dataset, machine_type)`). When we made that change, the implicit unknown-field check disappeared because there are no leftover keys to inspect when params are explicit. The validation here was added to try to preserve that but I do agree this should be tackled separately.

expertise/service/utils.py
cagostino: based on what I can find, we should be able to use this : https://docs.cloud.google.com/storage/docs/samples/storage-transfer-manager-upload-many

we'd still need a double for loop to get all the file names, but in this case it should distribute the uploads to multiple workers simultaneously instead of needing to be sequential

@tyler-dominguez do you think this approach makes sense instead ?

expertise/service/expertise.py
cagostino: upon further inspection the refactoring of the apirequest call site was implemented incorrectly and overzealously changed the upstream rather than just where it was being used. have changed this now

expertise/service/utils.py
```diff
@@ -28,6 +29,8 @@ def get_user_id(openreview_client):
 
     :returns id: The id of the logged in user
     """
+    if openreview_client is None:
```
melisabok: when is this happening?

expertise/service/utils.py
tyler-dominguez: Sounds like a good use case for Upload Many Objects, mitigates some of Carlos' concerns

expertise/service/expertise.py
```diff
@@ -864,22 +833,38 @@ async def worker_process(self, job, token):
...
+        try:
             cloud_id = self.cloud.create_job(
                 deepcopy(request),
                 job_id=job.id,
-                client=openreview_client_v2,
                 user_id=user_id,
-                machine_type=machine_type
+                machine_type=machine_type,
+                dataset_gcs_path=dataset_gcs_path,
+                vertex_id=vertex_id
```
melisabok: it looks like cloud_id and vertex_id are the same value, you are passing a value as a param and return it without any change. I would use one variable name, could this work?:

```
config.cloud_id = f"{job.id}-{int(time.time() * 1000)}"
self.cloud.create_job(
                config.cloud_id,
                deepcopy(request),
                job_id=job.id,
                client=openreview_client_v2,
                user_id=user_id,
                machine_type=machine_type
                machine_type=machine_type,
                dataset_gcs_path=dataset_gcs_path
)
```

expertise/create_dataset.py
```diff
@@ -362,12 +363,22 @@ def get_profiles(self, group_ids=None, reviewer_ids=None):
         if not all_members:
             return [], []
 
-        profiles_by_key = openreview.tools.get_profiles(
-            self.openreview_client_v2,
-            list(all_members),
-            with_publications=True,
-            as_dict=True
-        )
+        max_retries = 3
```
melisabok: can you revert these changes?

tests/test_expertise_cloud_service.py
```diff
@@ -349,14 +415,22 @@ def setup_job_mocks():
...
+        assert len(dataset_blobs) > 0, f"No dataset files found in bucket at {dataset_prefix}"
+        assert 'submissions.json' in dataset_blob_names, f"submissions.json not found in bucket. Found: {dataset_blob_names}"
+        assert any(name.startswith('archives/') for name in dataset_blob_names), f"No archive files found in bucket. Found: {dataset_blob_names}"
+
         # Upload test results to GCS
         metadata_blob = gcs_test_bucket.blob(f"{gcs_jobs_prefix}/{config.cloud_id}/metadata.json")
         metadata_blob.upload_from_string(json.dumps({"meta": "data"}))
 
-        scores_blob = gcs_test_bucket.blob(f"{gcs_jobs_prefix}/{config.cloud_id}/scores.jsonl")
+        scores_blob = gcs_test_bucket.blob(f"{gcs_jobs_prefix}/{config.cloud_id}/{config.cloud_id}.jsonl")
```
melisabok: revert this, let's keep using scores and scores_sparse

expertise/service/expertise.py
```diff
@@ -798,53 +796,24 @@ def set_client_v2(self, client_v2):
...
 
-    def compute_machine_type(self, client, job_id):
-        config = self.redis.load_job(job_id, get_user_id(client))
+    def compute_machine_type_from_dataset(self, config):
+        """Compute machine type from the already-created dataset on disk.
+
+        Reads submission_count from metadata.json written by execute_create_dataset().
+        Falls back to counting archive files if metadata is unavailable.
+        """
         if config.machine_type is not None:
```
melisabok: matchine_type is not computed before I believe, why do you need this? just remove it

expertise/service/expertise.py
```diff
@@ -798,53 +807,24 @@ def set_client_v2(self, client_v2):
...
+        metadata_path = os.path.join(config.job_dir, 'metadata.json')
+        if os.path.exists(metadata_path):
+            with open(metadata_path, 'r') as f:
+                metadata = json.load(f)
+            note_count = metadata.get('submission_count', 0) + metadata.get('archives_count', 0)
 
-            note_count += len(reduced_submissions)
+        self.logger.info(f"Machine type selection: {note_count} submissions in dataset")
 
         if note_count < self.server_config.get('PIPELINE_MEDIUM_THRESHOLD'):
```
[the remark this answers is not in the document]
melisabok: @cagostino can you check this comment?

expertise/service/expertise.py
```diff
@@ -864,24 +833,39 @@ async def worker_process(self, job, token):
...
+            self.logger.error(f"Error details: {traceback.format_exc()}")
+            config = self.redis.load_job(redis_id, user_id)
+            if config.status != JobStatus.ERROR:
+                self.update_status(config, JobStatus.ERROR, str(e))
+            raise e.with_traceback(e.__traceback__)
 
-            cloud_id = self.cloud.create_job(
+        # Upload dataset to GCS and submit VertexAI job (expertise only, Task 2)
+        config = self.redis.load_job(redis_id, user_id)
+        self.update_status(config, JobStatus.QUEUED)
```
melisabok: shouldn't be this status set when the job is sent to BulletMQ queue? why is being set after creating the dataset?

expertise/service/expertise.py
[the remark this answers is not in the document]
cagostino: claude suggests we double the current limits to account for this, i'll proceed with that and then we can change if we find it's causing issues

expertise/execute_pipeline.py
```diff
@@ -112,22 +129,28 @@ def run_pipeline(
...
         if working_dir is None:
             working_dir = f"/app/{job_id}"
         os.makedirs(working_dir, exist_ok=True)
 
+        validated_request = APIRequest({
+            'name': raw_request['name'],
+            'entityA': raw_request['entityA'],
+            'entityB': raw_request['entityB'],
+            **{k: raw_request[k] for k in ['model', 'dataset', 'machineType'] if raw_request.get(k) is not None}
+        })
```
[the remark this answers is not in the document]
cagostino: `raw_request` here carries service-internal fields (`token`, `baseurl_v2`, `gcs_folder`, `dump_archives`, etc.) that are popped out above before we construct the `APIRequest`. We handpick the API-request keys rather than passing the remaining dict directly, which is what bypasses `APIRequest`'s leftover-key check. Carlos already flagged in this PR that request validation should be redone properly with pydantic, so rather than patch in an ad-hoc allowlist/assertion here, we are tracking the rewrite as its own issue.

expertise/service/utils.py
```diff
@@ -943,9 +978,13 @@ def write_json_to_gcs(bucket_name, folder_path, file_name, data):
...
 
-        or_client = client if client else self.client
-        api_request = APIRequest(json_request)
-        valid_vertex_id = job_id + '-' + str(int(time.time() * 1000))
+        api_request = APIRequest({
+            'name': json_request['name'],
+            'entityA': json_request['entityA'],
+            'entityB': json_request['entityB'],
+            **{k: json_request[k] for k in ['model', 'dataset', 'machineType'] if json_request.get(k) is not None}
+        })
```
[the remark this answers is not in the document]
cagostino: Same reasoning as my reply on the `execute_pipeline.py` comment: `json_request` contains service-internal keys that have to be stripped before it can be passed into `APIRequest`, which is why we build the explicit dict here. `APIRequest` itself still does leftover-key validation on whatever it receives. Carlos already agreed in this PR that we should redo request validation properly with pydantic, so we are tracking that as its own issue rather than patching in an ad-hoc assertion here.

expertise/service/utils.py
```diff
@@ -972,13 +1011,17 @@ def write_json_to_gcs(bucket_name, folder_path, file_name, data):
...
+
+        # Select the per-tier pipeline; fall back to base name if tier mapping unavailable
+        tier_pipeline_name = getattr(self, 'pipeline_name_by_tier', {}).get(machine_type, self.pipeline_name)
 
         # Build PipelineJob kwargs and parameters
         job = aip.PipelineJob(
             display_name = valid_vertex_id,
-            template_path = f"https://{self.region}-kfp.pkg.dev/{self.project_id}/{self.pipeline_repo}/{self.pipeline_name}/{self.pipeline_tag}",
+            template_path = f"https://{self.region}-kfp.pkg.dev/{self.project_id}/{self.pipeline_repo}/{tier_pipeline_name}/{self.pipeline_tag}",
             job_id = valid_vertex_id,
```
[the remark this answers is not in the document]
cagostino: `pipeline_name_by_tier` is keyed by `SMALL_NAME`/`MEDIUM_NAME`/`LARGE_NAME` from the server config (utils.py:791-795), and `compute_machine_type_from_dataset` returns one of those same three values pulled from the same config (expertise.py:815-819). The two sources are in sync by construction, so the fallback branch is unreachable in practice — adding a `ValueError` here would be guarding an impossible condition. Leaving as-is.

expertise/service/expertise.py
```diff
@@ -798,53 +796,24 @@ def set_client_v2(self, client_v2):
...
-                paper_venueid=paper_venueid,
-                paper_content=paper_content,
-                submission_groups=submission_groups
-            )
+        note_count = 0
+        metadata_path = os.path.join(config.job_dir, 'metadata.json')
+        if os.path.exists(metadata_path):
+            with open(metadata_path, 'r') as f:
+                metadata = json.load(f)
+            note_count = metadata.get('submission_count', 0) + metadata.get('archives_count', 0)
```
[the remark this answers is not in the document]
cagostino: Resolved by removing the nonexistent fallback. `execute_create_dataset` always runs before `compute_machine_type_from_dataset` and always writes `metadata.json`, so the `os.path.exists` check and the `note_count = 0` default were guarding a condition that cannot occur. Now we just open `metadata.json` directly and the docstring matches the behavior.

expertise/execute_pipeline.py
```diff
@@ -137,9 +160,11 @@ def run_pipeline(
             for field in path_fields:
                 config.model_params[field] = working_dir
 
-        # Create Dataset and Execute Expertise
-        print('Creating dataset and executing expertise')
-        execute_create_dataset(client_v2, config.to_json())
+        if dataset_gcs_path:
+            print(f'Downloading pre-created dataset from {dataset_gcs_path}')
+            download_dataset_from_gcs(dataset_gcs_path, working_dir)
```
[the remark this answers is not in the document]
cagostino: Not doing this — Melisa confirmed `dataset_gcs_path` should stay optional so the expertise step can still be run locally against a pre-created dataset in `working_dir`. Reverting the fail-fast.

tests/test_pipeline.py
```diff
@@ -459,7 +489,7 @@ def test_runtime_errors(mock_load_model_artifacts, mock_execute_expertise, openr
             'scoreComputation': 'avg'
         },
         "user_id": "openreview.net",
-        "token": pipeline_client.token,
```
melisabok, line 462: can you keep using pipeline_client here?

melisabok (approved):

expertise/service/utils.py
```diff
@@ -924,7 +931,32 @@ def _generate_vertex_prefix(api_request):
...
+        """Upload dataset files from config.job_dir to GCS. Returns the GCS path."""
+        job_dir = config.job_dir
+        folder_path = f"{self.jobs_folder}/{config.job_id}/dataset"
+
+        dataset_items = ['archives', 'submissions', 'submissions.json']
+        for item in dataset_items:
+            local_path = os.path.join(job_dir, item)
+            if os.path.isdir(local_path):
+                for root, dirs, files in os.walk(local_path):
+                    for fname in files:
```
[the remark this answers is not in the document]
cagostino: Addressed in d3646df — `upload_dataset` now uses `storage.transfer_manager.upload_many_from_filenames`, which parallelises the uploads instead of doing them one by one in three nested loops. I kept each file as its own blob rather than tarring so the Vertex container can stream selectively via `download_dataset_from_gcs` on the other side.

expertise/service/expertise.py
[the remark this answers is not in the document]
cagostino: Doubled in 91e8ebf — `PIPELINE_MEDIUM_THRESHOLD` is now 2000 and `PIPELINE_LARGE_THRESHOLD` is 10000 in `default.cfg` to account for the new counter summing A + B + archives.

expertise/build_pipeline.py
```diff
@@ -173,64 +172,82 @@ def execute_expertise_pipeline_op(
...
-        extra_headers={"description": args.kfp_description}
-    )
-    print(f"Pipeline uploaded: templateName='{templateName}', versionName='{versionName}'")
\ No newline at end of file
+    for pipeline_func, pipeline_name, package_path in pipelines:
+        compiler.Compiler().compile(pipeline_func=pipeline_func, package_path=package_path)
+
+        version_exists = False
+        try:
+            existing_tag = registry_client.get_tag(package_name=pipeline_name, tag=args.tag)
```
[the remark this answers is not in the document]
cagostino: Yes — `build_pipeline.py` now compiles and uploads three separate pipelines per deploy (`{base_name}-small`, `-medium`, `-large`), each with its own `version_exists` guard against the `RegistryClient`. The existing `dev-deployment` / `prod-deployment` workflows pick them all up automatically since they invoke `build_pipeline.py` once per run.

melisabok (approved):

carlosmondra (approved):

PR #354, opened 2026-04-10
Fix: Prevent Vertex AI pipeline jobs from hanging on HuggingFace downloads

melisabok:
### Summary
- Pipeline jobs were intermittently hanging when Vertex AI workers couldn't reach HuggingFace Hub to download model weights at runtime
- Remove unnecessary `server.py` import from model files that triggered the entire Flask/BullMQ/GCP SDK stack during pipeline execution
- Pre-cache all HuggingFace models in the Docker image so jobs never depend on HuggingFace availability
- Skip GCS artifact download (`load_model_artifacts`) for models that don't need it (e.g. `specter2+scincl`)

### Changes

**Dockerfile**
- Add `PYTHONUNBUFFERED=1` for real-time log output in containers
- Pre-download all HuggingFace models (`allenai/specter`, `allenai/specter2_aug2023refresh_base`, `malteos/scincl`) and the specter2 adapter during image build

**expertise/models/\*/specter.py, scincl.py**
- Remove module-level `from expertise.service.server import redis_embeddings_pool` — this was unused in `specter2_scincl` and caused the entire web service stack (Flask, BullMQ, GCP AI Platform SDK) to initialize during pipeline jobs
- In `multifacet_recommender/specter.py`, the import is now lazy (only when `use_redis=True`)
- Add `print` statements with `flush=True` around model loading steps for better log visibility

**expertise/execute_pipeline.py**
- Skip `load_model_artifacts()` (GCS download) for models that don't use the old AllenNLP specter artifacts (e.g. `specter2+scincl`, `specter2`, `scincl`)

**expertise/execute_expertise.py**
- Add logging around model module import and initialization

**expertise/dataset/core.py**
- Add `flush=True` to dataset loading prints

### Root cause
HuggingFace models were downloaded from the internet on every pipeline job. When HuggingFace Hub was unreachable from certain Vertex AI workers, `AutoModel.from_pretrained()` and `load_adapter()` would hang indefinitely with no timeout or error. Additionally, importing `server.py` at module level caused the Flask app and all its dependencies to initialize unnecessarily during pipeline execution, which could also hang depending on the worker's network state.

### DIFF
--- a/expertise/execute_pipeline.py
+++ b/expertise/execute_pipeline.py
@@ -1,5 +1,11 @@
 import argparse
 import os
+
+# Set HuggingFace timeouts (in seconds) before importing any model code.
+# Prevents jobs from hanging indefinitely if HuggingFace Hub is unreachable.
+os.environ.setdefault('HF_HUB_DOWNLOAD_TIMEOUT', '600')
+os.environ.setdefault('HF_HUB_ETAG_TIMEOUT', '30')
+
 import openreview
 import shortuuid
 import json
--- a/expertise/execute_pipeline.py
+++ b/expertise/execute_pipeline.py
@@ -108,12 +108,8 @@ def run_pipeline(
         _, bucket = load_gcs(destination_prefix)
         blob_prefix = '/'.join(destination_prefix.split('/')[3:])
 
-        model_type = raw_request.get('model', DEFAULT_CONFIG['model'])
-        if model_type in ('specter+mfr', 'specter', 'mfr'):
-            print(f'Loading model artifacts for {model_type}')
-            load_model_artifacts()
-        else:
-            print(f'Skipping GCS artifact download for {model_type}')
+        print('Loading model artifacts')
+        load_model_artifacts()
 
         print('Logging into OpenReview')
         client_v2 = openreview.api.OpenReviewClient(baseurl_v2, token=token)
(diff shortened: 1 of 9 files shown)


carlosmondra (approved):

haroldrubio (approved):
Changes look good, but are we still planning on storing some version of the models in the bucket and downloading it from our network?

tyler-dominguez (approved):
Good solution to avoid hf always being down when needed

expertise/execute_pipeline.py
```diff
@@ -108,8 +108,12 @@ def run_pipeline(
         _, bucket = load_gcs(destination_prefix)
         blob_prefix = '/'.join(destination_prefix.split('/')[3:])
 
-        print('Loading model artifacts')
-        load_model_artifacts()
+        model_type = raw_request.get('model', DEFAULT_CONFIG['model'])
+        if model_type in ('specter+mfr', 'specter', 'mfr'):
```
cagostino: double checking here: is model_type a dict with param 'name' or will it be the model string directly? this was the main thing that Claude flagged but I don't think this is occurring. commenting here just in case it is worth double checking

cagostino (approved):
looks good,

expertise/execute_pipeline.py
melisabok: thanks for flagging this, I reverted the change and I will do it in another PR so I can test it. I noticed that we are copying many files that we don't use so I wanted to avoid that.

PR #360, opened 2026-04-21
count publications correctly in archives count

cagostino:


expertise/create_dataset.py
melisabok, line 759: I believe you can set archives_count here as well.

melisabok:
The changes look good, there are several tests validating that `assert metadata['submission_count'] == 2` is correct, could you do the same for archives_count?

melisabok (approved):

PR #366, opened 2026-04-27
Status update during dataset creation; tar dataset for upload/download

### LINKED ISSUE #364: Use different state than "queued" for jobs that are creating the dataset
When a job is creating the dataset, its status is "Queued". We should use a different state to distinguish between jobs that are pending in the queue and jobs that are creating the dataset and not yet in VertexAI.

### LINKED ISSUE #365: Compress dataset into a single archive before upload
Dataset transfer is slow for large jobs because `archives/` is split across one file per reviewer; each one is a separate transfer roundtrip. Tarring the dataset before upload and extracting on the other side collapses this to a single transfer.

### Upload side

`GCPInterface.upload_dataset()`:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/expertise/service/utils.py#L949-L977

Replace the `os.walk` collection and `transfer_manager.upload_many_from_filenames` call with a single tarball upload. Tar `archives/`, `submissions/`, `submissions.json`, `metadata.json` from `config.job_dir` into one `dataset.tar.gz` and upload that.

### Download side

`download_dataset_from_gcs()`:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/expertise/execute_pipeline.py#L60-L73

Replace the `list_blobs` enumeration and per-blob download loop with a single download of `dataset.tar.gz` followed by extraction into `local_dir`. Extracted layout matches what `execute_expertise` already expects.

### Tests to update

`test_upload_dataset`, `test_upload_dataset_with_vertex_id`, `test_upload_dataset_empty_directory`:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/tests/test_gcp_interface.py#L374-L510

Update assertions: a successful upload should produce a single `dataset.tar.gz` blob, not N per-file blobs. Add a round-trip test that tars a fixture directory, untars it, and asserts the resulting layout matches the input.

`test_run_pipeline` — assertion on per-archive blobs at L158:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/tests/test_pipeline.py#L46-L159

`test_run_pipeline_gcsdir` — assertion on per-archive blobs at L266:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/tests/test_pipeline.py#L162-L267

`test_run_pipeline_group` — assertion on per-archive blobs at L371:

https://github.com/openreview/openreview-expertise/blob/2d7bcb283b31655764fcba95bb2034658aed049d/tests/test_pipeline.py#L270-L372

In each, replace the `list_blobs(prefix=f"{prefix}archives/")` assertion with one that verifies a single `dataset.tar.gz` was produced (or staged, for the gcsdir variant).

cagostino:
### Title
Status update during dataset creation; tar dataset for upload/download

### Summary
- Set job status to `Fetching Data` while the worker builds the dataset, so users no longer see `Queued` for the entire dataset-creation window.
- Package the dataset into a single `dataset.tar.gz` before upload and extract on the Vertex side, replacing N per-reviewer object transfers with one.

Closes #364
Closes #365

### Motivation
Two symptoms in the current job flow:

1. Jobs display `Queued` while the worker is actively fetching data and building the dataset. `JobStatus.FETCHING_DATA` already exists with a fitting description but is never set — the worker goes straight from enqueue (`QUEUED`) to the polling loop (which only flips to `RUN_EXPERTISE` once Vertex transitions to RUNNING).
2. The dataset is uploaded and downloaded as one GCS object per reviewer. For large groups, per-object roundtrip overhead dominates the transfer phase even though total byte volume is modest.

### Changes

**`expertise/service/expertise.py` — status update before dataset creation**
- In `ExpertiseCloudService.worker_process`, call `self.update_status(config, JobStatus.FETCHING_DATA)` immediately before `execute_create_dataset(...)`. Status remains `FETCHING_DATA` through dataset creation and upload until the polling loop sees Vertex transition out of `PENDING`/`QUEUED`.

**`expertise/service/utils.py` — single-tarball upload**
- `GCPInterface.upload_dataset()` packages `archives/`, `submissions/`, `submissions.json`, `metadata.json` from `config.job_dir` into a single `dataset.tar.gz` (stdlib `tarfile` + `gzip`) and uploads that one blob.
- Returns the full GCS path to the tarball instead of the folder prefix. `transfer_manager` import is no longer needed here.

**`expertise/execute_pipeline.py` — single-tarball download**
- `download_dataset_from_gcs()` downloads `dataset.tar.gz` and extracts into `local_dir`. The extracted layout matches what `execute_expertise` already expects — no downstream changes.

**Tests**
- `tests/test_gcp_interface.py`:
  - `test_upload_dataset`, `test_upload_dataset_with_vertex_id`, `test_upload_dataset_empty_directory` updated to assert a single `dataset.tar.gz` blob is produced (or no upload when the directory is empty).
  - New round-trip test: tar a fixture directory via `upload_dataset`, untar via `download_dataset_from_gcs`, assert the resulting layout matches the input.
- `tests/test_expertise_cloud_service.py`:
  - New test for the status sequence — submit a job, intercept `execute_create_dataset` to assert the persisted status is `FETCHING_DATA` at that point, and that the status was `QUEUED` immediately after `start_expertise`.


melisabok (approved):

expertise/service/config/default.cfg
```diff
@@ -34,7 +34,7 @@ MEDIUM_NAME = 'medium'
 LARGE_NAME = 'large'
 
 PIPELINE_MEDIUM_THRESHOLD = 2000
-PIPELINE_LARGE_THRESHOLD = 10000
```
melisabok, line 37: I believe medium should increase as well

expertise/service/config/default.cfg
```diff
@@ -34,7 +34,7 @@ MEDIUM_NAME = 'medium'
...
+PIPELINE_LARGE_THRESHOLD = 40000
```
melisabok: large should be used for large venues: ICLR, NeurIPs, ICML, they usually have 20k submissions x 20k reviewers where each reviewer can have ~20 publications.

expertise/service/expertise.py
```diff
@@ -876,6 +881,9 @@ async def worker_process(self, job, token):
                     self.logger.info(f"INFO: after status check")
                     # Only update non-stale status
                     if config.status != status['status'] or config.description != status['description']:
+                        # Vertex reports QUEUED/INITIALIZED while we're still fetching data — skip regression
+                        if config.status == JobStatus.FETCHING_DATA and status['status'] in (JobStatus.QUEUED, JobStatus.INITIALIZED):
+                            continue
```
carlosmondra, line 887: I think you need to sleep here with asyncio, otherwise you will block the thread.

expertise/service/utils.py
```diff
@@ -947,33 +948,35 @@ def _generate_vertex_prefix(api_request):
...
+            tarball_path = tmp.name
+        try:
+            with tarfile.open(tarball_path, 'w:gz') as tar:
+                for item in items_to_pack:
+                    tar.add(os.path.join(job_dir, item), arcname=item)
+            self.bucket.blob(blob_name).upload_from_filename(tarball_path)
+            self.logger.info(f"Uploaded dataset tarball to {dataset_gcs_path}")
+        finally:
+            if os.path.exists(tarball_path):
+                os.remove(tarball_path)
```
carlosmondra, line 985: Do we need to remove the dataset_items too?

expertise/execute_pipeline.py
```diff
@@ -58,19 +60,23 @@ def download_from_gcs(gcs_path):
...
+        tarball_path = tmp.name
+    try:
+        bucket.blob(blob_name).download_to_filename(tarball_path)
+        print(f"Downloaded {blob_name} to {tarball_path}")
+        with tarfile.open(tarball_path, 'r:gz') as tar:
+            tar.extractall(local_dir)
+        print(f"Extracted dataset into {local_dir}")
+    finally:
+        if os.path.exists(tarball_path):
+            os.remove(tarball_path)
```
carlosmondra: You don't even need this. The instance is ephemeral, if I am following.
So I suppose we can let the exception bubble up

expertise/service/utils.py
cagostino: sure, but the compute_machine_type call after this relies on the metadata.json for computing that, but should be able to reorder them to accomplish this without interfering in that computation

carlosmondra: I just want to make sure that the local disk doesn't run out of memory

PR #367, opened 2026-05-01
testing out an interim embeddings caching mechanism for venues to re-…

cagostino:
…use embeddings produced by jobs run recently


melisabok (approved):

expertise/service/utils.py
```diff
@@ -41,6 +42,34 @@ def _get_required_field(req, superkey, key):
...
+    yields a prefix shared across requests for the same venue.
+    """
+    if not isinstance(api_request, dict):
+        return None
+    for entity_key in ('entityA', 'entityB'):
+        entity = api_request.get(entity_key) or {}
+        member = entity.get('memberOf')
+        if member:
+            return member.rsplit('/', 1)[0] if '/' in member else member
+        invitation = entity.get('invitation')
```
melisabok, line 59: can you include "id" to match TMLR?

melisabok: nvm, you are caching the publications

expertise/service/utils.py
```diff
@@ -41,6 +42,34 @@ def _get_required_field(req, superkey, key):
...
+        if invitation and '/-/' in invitation:
```
melisabok, line 60: the invitation is not necessary because we are caching user publications for now so only members of a group. Let's keep it for the future.

carlosmondra (approved):

PR #373, opened 2026-05-07
Fix `KeyError: 'weight'` in `all_scores` when reusing cached publication embeddings

melisabok:
## What this changes

When a request includes a `weightSpecification` (i.e. `venue_specific_weights=True`) and the pipeline reuses a `cached_pub2vec_*.jsonl` file produced by a prior job, the Vertex AI worker crashes with:

```
File "/app/openreview-expertise/expertise/models/specter2_scincl/specter.py", line 246, in load_emb_file
    weight_list.append(paper_data['weight'])
KeyError: 'weight'
```

The cached embedding lines were written without a `weight` field, but `all_scores` was reading the weight from the embedding line — so any cache hit broke the run.

The root cause is a design issue: an embedding only depends on title + abstract, but `weight` (which depends on the request and the publication's venue) was being persisted alongside it. Two requests for the same paper can legitimately want different weights, so caching them together is incorrect by construction. This PR decouples the two: weights are now read from the per-job metadata file (`*_reviewer_paper_data.json`, populated by `set_archives_dataset` for the current request), and embeddings stay request-agnostic.

### Files changed

- **[expertise/models/specter2_scincl/predictor.py](expertise/models/specter2_scincl/predictor.py)** — `_build_embedding_jsonl` no longer copies `weight` into the JSONL. Embedding files now contain only `paper_id` and `embedding`.
- **[expertise/models/specter2_scincl/specter.py](expertise/models/specter2_scincl/specter.py)** — in `all_scores`, when `venue_specific_weights` is on, build a `paper_id → weight` map by reading `specter_reviewer_paper_data.json` from the work dir. Pass that map into the inner `load_emb_file` instead of the old `load_weight=True` flag, which used to read `paper_data['weight']` off the embedding line. Defaults to `1.0` for any paper missing from the metadata. Also makes `set_archives_dataset` safe against publications missing `content.weight` — `publication['content']['weight']` → `publication.get('content', {}).get('weight', 1)`, matching the pattern already used in scincl.
- **[expertise/models/specter2_scincl/scincl.py](expertise/models/specter2_scincl/scincl.py)** — same `all_scores` change against `scincl_reviewer_paper_data.json`. (`set_archives_dataset` here already used the safe `.get` default.)
- **[tests/test_specter2scincl.py](tests/test_specter2scincl.py)** — adds `test_venue_specific_weights_with_weightless_cache`, a regression test that:
  1. Builds an `EnsembleModel` with `venue_specific_weights=True` and archives that carry per-publication weights.
  2. Runs `embed_publications` once to materialise embeddings, then strips them into `cached_pub2vec_specter.jsonl` and `cached_pub2vec_scincl.jsonl` with the `weight` field removed — exactly the shape produced by a prior job that ran without venue weights.
  3. Re-runs the pipeline; the cache hits, no `_batch_predict` is called, and `all_scores` must complete without `KeyError`.
  4. Asserts the resulting publication embedding lines no longer carry a `weight` key — guards against a future regression that re-couples the two.
- **[tests/test_expertise_service.py](tests/test_expertise_service.py)** — `test_request_expertise_with_valid_parameters` previously asserted weights had been written into `pub2vec_specter.jsonl` / `pub2vec_scincl.jsonl`. That was the *old* contract; after the refactor weights live in `*_reviewer_paper_data.json`. Updated the two assertion blocks (around lines 644 and 743) to read from `specter/specter_reviewer_paper_data.json` and `scincl/scincl_reviewer_paper_data.json` instead. Same coverage, new file.

## Why this slipped past existing tests

The bug only triggers when **cache reuse** and **`venue_specific_weights=True`** happen in the same run. Existing tests covered the two axes separately but never together:

| Test | Cache path | Venue weights |
|---|---|---|
| `test_request_expertise_with_valid_parameters` (`test_expertise_service.py`) | ❌ | ✅ |
| `test_cached_embeddings_produce_identical_results` (`test_specter2scincl.py`) | ✅ | ❌ |
| `test_load_cached_publication_embeddings` (`test_utils.py`) | ✅ (loader only) | ❌ |
| GCP cache tests (`test_gcp_interface.py`) | ✅ (upload/download only) | ❌ |
| **`test_venue_specific_weights_with_weightless_cache` (new)** | ✅ | ✅ |

That's why the failure only surfaced in production (NeurIPS 2026 Reviewers job, where prior venue jobs had warmed `cached_pub2vec_*.jsonl`) instead of CI. The new test fills exactly that cell.

## Why not just inject the weight into cached lines

That was the first fix: re-parse each cached line and patch `weight` from the current `paper_data` before writing it out. It works, but it's a band-aid — it preserves the bad coupling between embeddings (a function of content) and weights (a function of the request). Reading weights from the metadata file at scoring time eliminates the bug class instead of patching one symptom, and `set_archives_dataset` already writes that metadata file unconditionally on every run.

## Impact on existing caches

`cached_pub2vec_*.jsonl` files written by older builds had a `weight` field; new builds won't. Both are accepted on the read path — `_load_cached_publication_embeddings` only pulls `paper_id` and `embedding`, and the new `all_scores` path doesn't look at the embedding line for weight at all. No cache invalidation needed.

## Test plan

- [x] `pytest tests/test_specter2scincl.py::test_venue_specific_weights_with_weightless_cache` passes.
- [x] `pytest tests/test_specter2scincl.py::test_cached_embeddings_produce_identical_results` still passes (cached embeddings continue to round-trip).
- [x] `pytest tests/test_utils.py::test_load_cached_publication_embeddings` still passes.
- [ ] `pytest tests/test_expertise_service.py::TestExpertiseService::test_request_expertise_with_valid_parameters` passes on CI (the assertion paths were updated to point at the metadata files).
- [ ] Re-run the failing Vertex AI job (NeurIPS 2026 Reviewers, `weightSpecification` set) and confirm `all_scores` completes.
- [ ] One run with `weightSpecification` set + warm `cached_pub2vec_*.jsonl` from a prior venue job.
- [ ] One run without `weightSpecification` (sanity: weight tensor stays empty, no metadata read attempted).


carlosmondra (approved):

tests/test_specter2scincl.py
```diff
@@ -533,3 +533,90 @@ def test_cached_embeddings_produce_identical_results(tmp_path):
...
+    model.embed_publications(specter_pub_path, scincl_pub_path)
+    model.embed_submissions(specter_sub_path, scincl_sub_path)
+
+    # Sanity: cache was actually consumed (no 'weight' key in publication lines).
+    for path in (specter_pub_path, scincl_pub_path):
+        for line in path.read_text().splitlines():
+            assert "weight" not in json.loads(line), (
+                f"Publication embedding lines should not carry 'weight' anymore: {path}"
+            )
+
```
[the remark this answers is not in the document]
melisabok: fixed

expertise/service/utils.py
```diff
@@ -59,6 +60,9 @@ def extract_venue_key(api_request):
         invitation = entity.get('invitation')
         if invitation and '/-/' in invitation:
             return invitation.split('/-/')[0]
+        venueid = entity.get('withVenueid')
+        if venueid:
+            return venueid
```
melisabok: this part is tricky because when the submission are under review the venueid is `NeurIPS.cc/2026/Conference/Submission` that is not the same as the venueid that you get from the memberOf `NeurIPS.cc/2026/Conference`. 

One solution is to drop the last item after splitting by '/' but it not 100% secure. There could be other venueids that won't work but it does work for NeurIPS.

tests/test_utils.py
```diff
@@ -169,6 +169,14 @@ def test_artifacts_for_model_unknown_model_falls_back_to_all():
     ({'entityA': {}, 'entityB': {}}, None),
     # invitation without /-/ falls through to None
     ({'entityA': {'invitation': 'no-divider'}, 'entityB': {}}, None),
+    # entityB Note / withVenueid — split jobs use this when invitation is absent
+    ({'entityA': {'type': 'Group', 'reviewerIds': ['~A1', '~B1']},
+      'entityB': {'type': 'Note', 'withVenueid': 'ICLR.cc/2026/Conference'}},
```
melisabok: try passing ICLR.cc/2026/Conference/Submission here

cagostino (approved):

PR #382, opened 2026-05-11
adding tests for sparse score generation coverage.     

cagostino:
Contract tests cover:
- Top-k per submission and per reviewer, union'd (typical case)
- `sparse_value=1` argmax-only case
- `sparse_value` larger than groups → no-op
- 1×1 degenerate
- Output ordering (grouped by submission desc, score desc within group) — downstream JSONL streaming depends on this
- Return value matches CSV
- Every submission_id and reviewer_id appears
- Random 6×5 matrix at k ∈ {1, 2, 3, 5, 10}
- Tie behavior

### DIFF
--- a/tests/test_specter2scincl.py
+++ b/tests/test_specter2scincl.py
@@ -191,43 +191,54 @@ def test_sparse_scores(tmp_path, create_specncl):
         assert profile_id.startswith('~')
         assert score >= 0 and score <= 1
 
-    # Contract check on the actual sparse CSV produced by the pipeline.
-    # Builds a reference sparse selection (top-k per submission ∪ top-k per
-    # reviewer) directly from the full score list returned by all_scores,
-    # and asserts the on-disk _sparse.csv matches it. This verifies that
-    # the sparse CSV is a correct sparsification of the full scores
-    # independently of which implementation produced it — when the WIP
-    # matrix/topk-based sparsifier replaces generate_sparse_scores, this
-    # assertion still pins the behavior.
-    sparse_value = config['model_params']['sparse_value']
-    by_axis_top_k = set()
+    # End-to-end ordering check on the sparse CSV: for each reviewer, the
+    # papers selected for that reviewer (sorted by score desc) must match
+    # the top-`sparse_value` papers for that reviewer from the full scores;
+    # for each submission, vice versa. Runs whichever sparsifier the
+    # pipeline calls (current generate_sparse_scores or a future
+    # matrix/topk-based replacement), so it catches regressions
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -238,4 +238,218 @@ def test_load_cached_publication_embeddings(tmp_path):
...
+    matrix = [
+        [0.90, 0.10, 0.50, 0.60],
+        [0.20, 0.80, 0.30, 0.40],
+        [0.70, 0.50, 0.20, 0.15],
+    ]
+    test_ids = ["subA", "subB", "subC"]
+    reviewer_ids = ["R1", "R2", "R3", "R4"]
+    out = tmp_path / "sparse.csv"
+    generate_sparse_scores(_flatten_matrix(matrix, test_ids, reviewer_ids), 2, out)
+    assert _parse_sparse_csv(out) == _expected_rounded(matrix, test_ids, reviewer_ids, 2)
```
melisabok: I wonder if the assertion can be "hardcoded" here instead of using _expected_rounded to make sure we do not have bugs in the test code.

tests/test_specter2scincl.py
```diff
@@ -189,6 +191,44 @@ def test_sparse_scores(tmp_path, create_specncl):
...
+    on_disk = set()
+    with open(scores_path.joinpath(config['name'] + '_sparse.csv')) as f:
+        for line in f:
+            line = line.strip()
+            if not line:
+                continue
+            s, r, sc = line.split(',')
+            on_disk.add((s, r, float(sc)))
+
+    assert on_disk == by_axis_top_k, (
```
melisabok: what are you checking here?

tests/test_utils.py
```diff
@@ -238,4 +238,218 @@ def test_load_cached_publication_embeddings(tmp_path):
     assert jsonl_lines == {
         "p1": '{"paper_id": "p1", "embedding": [0.1, 0.2]}\n',
         "p2": '{"paper_id": "p2", "embedding": [0.3, 0.4]}\n',
-    }
\ No newline at end of file
+    }
+
+
+# ---------------------------------------------------------------------------
+# generate_sparse_scores — output contract
```
melisabok, line 245: I don't think these tests will help to verify the new changes. It is going to be another utils function that uses the topK call.

I recommend making sure an end to end expertise call is returning the sparse scores in the same order. We do not need to assert the exact score but we can make sure the papers are ordered correctly for each reviewers and viceversa

melisabok (approved):

PR #387, opened 2026-05-13
New `/expertise/metadata` endpoint

melisabok:
## Summary

Adds `GET /expertise/metadata?jobId=<id>`, which returns the dataset metadata for a completed job: submission/archive counts plus the lists of profiles and publications that couldn't be matched (`no_profile`, `no_publications`, `no_profile_submission`).

The previous PR added `?format=csv` to `/expertise/results` to skip JSON parsing for large jobs. CSV has no place for the metadata envelope, so clients that switch to CSV currently lose access to the "missing" lists that venue organizers need to surface in the UI.

## Why a separate endpoint

Mixing a `format=metadata` mode into `/expertise/results` would make the "results" route return no results. A dedicated route is one blob read, self-documenting, and can be called independently of how scores are fetched. The response is the raw `metadata.json` dict — no curated projection, so anything `execute_create_dataset` writes shows up automatically.

## API

`GET /expertise/metadata?jobId=X` → `application/json`:

```json
{
  "submission_count": 2,
  "archives_count": 10,
  "no_profile": [...],
  "no_publications": [...],
  "no_publications_count": 0,
  "no_profile_submission": [...]
}
```

Same auth (`require_auth`) and error mapping as the other routes (400 / 403 / 404 / 500). Requires job status `Completed`.

## Changes

- [`service/expertise.py`](expertise/service/expertise.py) — `ExpertiseService.get_expertise_metadata` reads `metadata.json` from `config.job_dir`; `ExpertiseCloudService.get_expertise_metadata` validates status and delegates to the GCP interface.
- [`service/utils.py`](expertise/service/utils.py) — `GCPInterface.get_job_metadata` mirrors the auth pattern in `get_job_results`, then downloads the top-level `metadata.json` blob (excluding the one packed inside `dataset/`, which is the input artifact).
- [`service/routes.py`](expertise/service/routes.py) — new blueprint route, standard try/except error mapping.

## Tests

- `tests/test_expertise_service.py` — three new tests: happy path (asserts counts + that the missing lists are present), unknown job (404), missing `jobId` (400).
- `tests/test_expertise_cloud_service.py::test_journal_results` — extended to also hit `/expertise/metadata` and assert it returns the metadata blob uploaded to the test GCS bucket.

## Migration notes

No changes to existing routes. `/expertise/results` still returns the same JSON envelope by default — clients on `?format=csv` should make a parallel call to `/expertise/metadata` for the missing-profile / missing-publication lists.

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -770,6 +770,28 @@ class ExpertiseService(BaseExpertiseService):
 
         return result
 
+    def get_expertise_metadata(self, user_id, job_id):
+        """
+        Gets the dataset metadata for a given job (submission/archive counts,
+        missing profiles and publications). Lighter than get_expertise_results
+        since it doesn't read the score file.
+        """
+        config = self.redis.load_job(job_id, user_id)
+
+        if config.status != JobStatus.COMPLETED:
+            raise openreview.OpenReviewException(
+                f"Metadata not available - status: {config.status} | description: {config.description}"
+            )
+
+        metadata_path = os.path.join(config.job_dir, 'metadata.json')
+        if not os.path.isfile(metadata_path):
+            raise openreview.OpenReviewException(
+                f"Metadata file not found for job {job_id}"
+            )
+
+        with open(metadata_path, 'r') as f:
+            return json.load(f)
+
 class ExpertiseCloudService(BaseExpertiseService):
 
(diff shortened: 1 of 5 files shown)


cagostino (approved):

carlosmondra (approved):
At some point I think we should retrieve this data from Redis instead of the bucket.

PR #392, opened 2026-06-03
make bullmq single source of truth for status of jobs

### LINKED ISSUE #284: Single source of truth
Currently there the status of a job is stored in more than one place. I propose the queue to be the source of truth when it comes to checking the state of a job.
We need to discuss what's stored in Redis and GCP at the moment regarding the state of the job.

cagostino:
Closes #284.

Status endpoints were reading from config.status in Redis, which only updates at a few points in worker_process(). A job could be active or already completed/failed in BullMQ while the API still reported QUEUED from Redis. 

This PR makes BullMQ the primary source of truth when requesting job information and keeps Redis lookup as a fallback when jobs are no longer found in the queue (100 completed, last 30 days for failed) or the queue is not responding.

Status is retrieved via `GET /expertise/status?jobId=...`, `get_expertise_status state`, `_get_job_status_from_queue()` which reads `queue.getJobState(job_id)` the BullMQ `Job` object itself when completed/active. The worker writes sub_status and result into job.data as it progresses, so the API can
distinguish `FETCHING_DATA`, `RUN_EXPERTISE`, `COMPLETED`, `DATA_ERROR`, and `ERROR` writes.

`get_expertise_results()` and `get_expertise_metadata()` use the same queue-first lookup, and `del_expertise_job()` now checks BullMQ before allowing deletion. 

`get_expertise_status_all` needs to be updated to check the queue
https://github.com/openreview/openreview-expertise/blob/5f36f64f83663d81e402d1cc7564bf60b608fd70/expertise/service/expertise.py#L352


carlosmondra:
So are we still going to use Redis as fallback? Can you explain why we need to do that? Maybe I'm missing something. Also, is it the case now that once the job completes, the job data is removed from Redis (not BullMQ)?

expertise/service/expertise.py
```diff
@@ -184,11 +176,83 @@ def update_status(self, config, new_status, desc=None):
...
+        is no longer in the queue (archived) or the query times out.
+        """
+        import concurrent.futures
+
+        try:
+            future = asyncio.run_coroutine_threadsafe(
+                self.queue.getJobState(job_id),
+                self.queue_loop
+            )
+            state = future.result(timeout=0.1)
```
carlosmondra: Why do we need this timeout?

expertise/service/expertise.py
```diff
@@ -184,11 +176,83 @@ def update_status(self, config, new_status, desc=None):
...
+            data['error'] = error
+        job.data = data
+        future = asyncio.run_coroutine_threadsafe(
+            self.queue.scripts.updateData(job.id, data),
+            self.queue_loop
+        )
+        await asyncio.wrap_future(future)
 
-        self._save_config(config)
+    def _get_job_status_from_queue(self, job_id):
```
carlosmondra: Why do you need to make 2 calls to the queue?
1. To get the state of the job
2. To get the data of the job
Couldn't you just get the state from the job object?

expertise/service/expertise.py
```diff
@@ -184,11 +176,83 @@ def update_status(self, config, new_status, desc=None):
...
+        except concurrent.futures.TimeoutError:
+            self.logger.warning(f"Timeout fetching job {job_id} from queue")
+        except Exception as e:
+            self.logger.warning(f"Failed to fetch job {job_id} from queue: {e}")
+
+        data = job.data if job else {}
+        status = data.get('status')
+        description = data.get('description')
+        if status is not None:
+            return status, (description or descriptions.get(status, ''))
```
carlosmondra, line 216: If status in the job is going to take precedence, why do you first try to get the state? You are wasting a call to Redis.

expertise/service/expertise.py
```diff
@@ -302,6 +367,37 @@ def check_result():
...
+
+        :param user_id: The ID of the user accessing the data
+        :type user_id: str
+
+        :param job_id: ID of the specific job to look up
+        :type job_id: str
+
+        :returns: A dictionary with the key 'results' containing a list of job statuses
+        """
+        config = self.redis.load_job(job_id, user_id)
```
carlosmondra: Do we still need this? The goal of doing this is to also remove the dependency on Redis.

expertise/service/expertise.py
```diff
@@ -630,7 +727,7 @@ def start_expertise(self, request, client):
                         'type': 'exponential', # Exponential backoff: 2 ^ attempts * delay milliseconds
                     },
                     'removeOnComplete': {
-                        'count': 100,
+                        'age': 2592000
```
carlosmondra: I'm wondering if this should be in the config file

expertise/service/expertise.py
carlosmondra: Is this for backward compatibility?

carlosmondra:
Is the plan to make a second PR to remove the remaining parts where you are calling to Redis?
I saw that you updated the time that job lives in the queue after it has completed. @melisabok, is it a problem if the job data disappears after this amount of time?

melisabok:
@melisabok, is it a problem if the job data disappears after this amount of time?

the time that stays in the queue is for 30 days or the top 100?

expertise/service/config/default.cfg
```diff
@@ -9,6 +9,8 @@ REDIS_PORT = 6379
 REDIS_CONFIG_DB = 10
```
melisabok, line 9: can this config be removed?

expertise/service/expertise.py
```diff
@@ -184,124 +170,221 @@ def update_status(self, config, new_status, desc=None):
...
+        if job is not None:
+            data = job.data
+            status = data.get('status')
+            description = data.get('description')
+            if status is not None:
+                return status, (description or descriptions.get(status, ''))
+
+        return None, None
+
+    def _get_job_from_queue(self, job_id):
```
melisabok: what is the difference between this code and the code above?

expertise/service/config/default.cfg
cagostino: these are still being used by the queue when it sets up 
https://github.com/openreview/openreview-expertise/blob/89de55cbffbd52c963c5b9128123c8e0b6a49678/expertise/service/expertise.py#L86

expertise/service/expertise.py
cagostino: this was updated so that job.fromId is only called in get job from queue, and now get job status from queue calls that and parses the status.

expertise/service/expertise.py
cagostino: redis no longer being called in this flow so resolving this

melisabok (approved):

expertise/service/expertise.py
cagostino: only one place calls it now https://github.com/openreview/openreview-expertise/pull/392/changes#diff-55a4d539fa7a3ae8514888fef2b7aa05a43f0c881d8206dd47b90023559e8d0aR192

expertise/service/expertise.py
cagostino: removed along with the Redis Database class

expertise/service/expertise.py
```diff
@@ -1000,7 +1072,7 @@ def start_expertise(self, request, client):
                         'type': 'exponential', # Exponential backoff: 2 ^ attempts * delay milliseconds
                     },
                     'removeOnComplete': {
-                        'count': 100,
+                        'age': 2592000
```
carlosmondra, line 1075: I thought we said that this was going to go in the config

cagostino: it is. https://github.com/openreview/openreview-expertise/pull/392/changes#diff-a2da9de9c400e8f238e95d73f64458a2df49e875e75ad58284e484e6039b5d6b
https://github.com/openreview/openreview-expertise/pull/392/changes#diff-55a4d539fa7a3ae8514888fef2b7aa05a43f0c881d8206dd47b90023559e8d0aR60

carlosmondra (approved):

PR #393, opened 2026-06-26
Fix bullmq Worker asyncio loop mismatch

cagostino:
- Instantiate Worker inside the daemon thread's event loop instead of on
the main gunicorn thread. BullMQ 2.11.0 creates internal asyncio.Futures
during Worker construction; when the Worker was constructed on one loop
but run() was scheduled on another, those Futures attached to the wrong
loop causing RuntimeError.
- Set autorun: False on Worker settings
- Move Worker instantiation into start_worker_in_thread()
- Use run_until_complete(worker.run()) instead of run_forever()
- Make close() defensive against self.worker = None

### DIFF
--- a/expertise/service/expertise.py
+++ b/expertise/service/expertise.py
@@ -84,25 +84,21 @@ class BaseExpertiseService:
         )
         self.start_queue_in_thread()
 
-        worker_settings = {
+        self.worker_settings = {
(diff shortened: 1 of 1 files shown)


melisabok (approved):

carlosmondra (approved):
OK, let's try it out

PR #395, opened 2026-07-01
Embedding Cache Integration

### LINKED ISSUE #205: Expertise API/Image: Cache or create a store of embeddings
We previously cached embeddings in an addon module for Redis but ran into some issues with it failing to fetch.

Now that we are executing the model serverless, we need a way for an expertise container to pull embeddings from, most likely Google Cloud, into its local filesystem.

A model-agnostic implementation may be difficult since the embedding step happens at the model level. This might have to happen in the `create_dataset` step and only embed publications in the filesystem, and merge the 2 embedding files.

cagostino:
Closes #205.

## Overview of changes across pipeline

1. `worker_process` creates dataset and uploads tarball of job data, cached pub2vec jsons are no longer used. ([`expertise/service/expertise.py:835`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/service/expertise.py#L835))
2. `run_pipeline` downloads dataset, no longer loads the cached embeddings from jsons ([`expertise/execute_pipeline.py:94`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_pipeline.py#L94))
3. Extracts paper IDs and mdate from archives/`*.jsonl`, submissions/`*.jsonl`, and `submissions.json` ([`expertise/execute_pipeline.py:197`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_pipeline.py#L197))
4. Models mapped to cache keys and cache prefix inferred ([`expertise/execute_pipeline.py:257`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_pipeline.py#L257)) for dev/prod cache separation
5. Global parquet cache queried via `GlobalEmbeddingsCache.get_embeddings_for_models` ([`expertise/execute_pipeline.py:265`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_pipeline.py#L265)), filtering stale embeddings by mdate ([`expertise/embeddings_cache.py:30`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/embeddings_cache.py#L30))
6. Cached embeddings are passed as a single `cached_embeddings` dict to `execute_expertise` ([`expertise/execute_expertise.py:14`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_expertise.py#L14)). The dict maps model name (`specter`, `scincl`) to `{paper_id: embedding}`. `execute_expertise` sets `predictor.cached_embeddings = (cached_embeddings or {}).get('specter', {})` (or `'scincl'`) before calling `embed()`
7. Predictors skip inference for cached papers, passing only those without embeddings in cache to batch predictors. [`embed()`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/models/specter2_scincl/predictor.py#L15) loads the metadata JSON file, checks `self.cached_embeddings` for hits ([`predictor.py:19-27`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/models/specter2_scincl/predictor.py#L19-L27)), computes only missing papers via `_batch_predict`, and returns the full embedding dict (cached + newly computed). Old duplicate `embed_publications()` / `embed_submissions()` methods are removed from predictors; `EnsembleModel.embed_publications()` and `.embed_submissions()` are removed from both [`specter2_scincl/ensemble.py`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/models/specter2_scincl/ensemble.py#L9) and [`multifacet_recommender/ensemble.py`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/models/multifacet_recommender/ensemble.py#L8). `execute_expertise` calls each component predictor's `.embed()` directly for computing submission and publication embeddings and assigns the result to `predictor.publication_embeddings` / `predictor.submission_embeddings` before `all_scores()`
8. Runs expertise, writes scores. Each predictor initializes `self.cached_embeddings = {}` in `__init__`. `_batch_predict` returns `[{'paper_id': ..., 'embedding': [...]}]` dicts instead of pre-serialized JSONL strings; `embed()` handles JSONL serialization in one place
9. Dumps embeddings JSONL files to job folder for backwards compatibility
10. Appends newly computed embeddings back to global cache via `_append_embeddings_to_global_cache` ([`expertise/execute_pipeline.py:381`](https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_pipeline.py#L381)). Cache is hive-partitioned by `model`, `year_month`, and `job_id`. `embedding_date` is stored as `timestamp[ms]`. `execute_expertise` filters the returned `new_embeddings` dict so it only contains paper IDs that were **not** already in `self.cached_embeddings`, preventing the global cache append from rewriting already-cached data


expertise/service/expertise.py
```diff
@@ -917,18 +862,16 @@ async def worker_process(self, job, token):
         config.cloud_id = f"{job.id}-{int(time.time() * 1000)}"
         machine_type = self.compute_machine_type_from_dataset(config)
         self.logger.info(f"Machine type for {redis_id}: {machine_type}")
-        try:
-            self._populate_publication_cache(config, request, job)
-        except Exception as e:
-            self.logger.warning(f"Embedding cache lookup failed for {redis_id}: {e}")
+
         asyncio.run_coroutine_threadsafe(job.log(f'Uploading dataset to gs://{self.cloud.bucket_name}/{self.cloud.jobs_folder}/{config.cloud_id}/dataset'), self.queue_loop)
```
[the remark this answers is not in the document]
cagostino: updating pr description

expertise/models/specter2_scincl/specter.py
```diff
@@ -181,49 +186,69 @@ def set_submissions_dataset(self, submissions_dataset):
...
+        for batch_data in tqdm(self._fetch_batches(remaining, self.batch_size), desc='Embedding Subs', total=int(len(remaining.keys())/self.batch_size), unit="batches"):
+            for item in self._batch_predict(batch_data):
+                self.submission_embeddings[item['paper_id']] = item['embedding']
+                new_embeddings[item['paper_id']] = item['embedding']
+
+        if submissions_path:
+            with open(submissions_path, 'w') as f:
+                for pid, emb in self.submission_embeddings.items():
+                    f.write(json.dumps({'paper_id': pid, 'embedding': emb}) + '\n')
+        return new_embeddings
```
carlosmondra: This code is repeated several times across more than one file. Is there a way to define it just once?

expertise/create_dataset.py
```diff
@@ -613,8 +613,21 @@ def get_submissions_helper(self,
...
+                            return datetime.datetime.fromtimestamp(raw / 1000, tz=datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+                        except Exception:
+                            return None
+                    return str(raw)
+
+                paper_mdate = _iso_mdate(getattr(paper, 'mdate', None))
+
                 reduced_submissions[paper_id] = {
                     'id': paper_id,
+                    'mdate': paper_mdate,
```
carlosmondra: Why do you change the format of mdate? I don't think it's necessarily wrong, but we use unix timestamps with mdate everywhere.

expertise/execute_expertise.py
```diff
@@ -1,14 +1,21 @@
...
+import openreview
 from .create_dataset import OpenReviewExpertise
 from .dataset import ArchivesDataset, SubmissionsDataset, BidsDataset
 from .config import ModelConfig
 from .utils.utils import aggregate_by_group, generate_sparse_scores, generate_sparse_scores_from_matrix
 
+
 # Move run.py functionality to a function that accepts a config dict
-def execute_expertise(config):
+def execute_expertise(config, cached_publication_embeddings=None, cached_submission_embeddings=None):
```
melisabok: why do we need two caches? I don't think we need to distinguish between publications or submissions.

expertise/execute_expertise.py
```diff
@@ -53,15 +60,27 @@ def execute_expertise(config):
...
             if config['model_params'].get('use_redis', False):
                 publication_path = None
-            predictor.embed_publications(publications_path=publication_path)
-            predictor.embed_submissions(submissions_path=Path(config['model_params']['submissions_path']).joinpath('sub2vec.jsonl'))
-        predictor.all_scores(
-            publications_path=publication_path,
-            submissions_path=Path(config['model_params']['submissions_path']).joinpath('sub2vec.jsonl'),
-            scores_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '.csv')
-        )
+            if cached_publication_embeddings is not None or cached_submission_embeddings is not None:
```
melisabok: why do you need to check if the cache is None here, can you do it inside of each embed_submissions function? the if/else seem to be doing the same.

expertise/execute_expertise.py
```diff
@@ -118,18 +158,29 @@ def execute_expertise(config):
...
-            Path(config['model_params']['submissions_path']).joinpath('sub2vec.jsonl')
-        )
-        predictor.all_scores(
-            scincl_publication_path,
-            Path(config['model_params']['submissions_path']).joinpath('sub2vec.jsonl'),
-            Path(config['model_params']['scores_path']).joinpath(config['name'] + '.pt'),
-            p2p_path=Path(config['model_params']['scores_path']).joinpath(config['name'] + '_p2p' + '.json')
-        )
+        submissions_path = Path(config['model_params']['submissions_path']).joinpath('sub2vec.jsonl')
+        if cached_publication_embeddings is not None or cached_submission_embeddings is not None:
```
melisabok: same here, if/else are doing the same

expertise/create_dataset.py
cagostino: I had done so in the date since I was extracting year and month for the partitioning but keeping both in timestamp will be simpler

expertise/execute_pipeline.py
```diff
@@ -197,8 +185,117 @@ def run_pipeline(
...
+                cached_publication_embeddings = {cache_key: {} for cache_key in targets}
+                cached_submission_embeddings = {cache_key: {} for cache_key in targets}
+                total_cached = 0
+
+                embeddings_by_model = cache.get_embeddings_for_models(list(note_ids), targets, paper_mdates=paper_mdates)
+
+                for cache_key in targets:
+                    model_embeddings = embeddings_by_model.get(cache_key, {})
+                    cached_publication_embeddings[cache_key] = model_embeddings
+                    cached_submission_embeddings[cache_key] = model_embeddings
```
carlosmondra: Why do we need to different caches?

expertise/execute_pipeline.py
```diff
@@ -261,17 +356,21 @@ def run_pipeline(
...
         print("Dumping embeddings", flush=True)
         for emb_file in [d for d in os.listdir(config.job_dir) if '.jsonl' in d]:
             destination_blob = f"{blob_prefix}/{emb_file}"
             blob = bucket.blob(destination_blob)
             blob.upload_from_filename(os.path.join(config.job_dir, emb_file))
 
+        # Append newly computed embeddings to the global parquet cache so future
+        # jobs can reuse them without recomputation.
+        try:
+            _append_embeddings_to_global_cache(new_embeddings, blob_prefix, bucket)
```
carlosmondra: How long does this step take given that is happening in the vertex instance?

expertise/execute_pipeline.py
```diff
@@ -293,6 +392,100 @@ def run_pipeline(
...
+            )
+            if existing_table.num_rows == 0:
+                continue
+            latest_table = existing_table.group_by("paper_id").aggregate([("embedding_date", "max")])
+            latest_rows = latest_table.to_pydict()
+            latest_embedding_date = {
+                pid: edate for pid, edate in zip(latest_rows["paper_id"], latest_rows["embedding_date_max"])
+                if edate is not None
+            }
+            if latest_embedding_date:
```
carlosmondra: Maybe I'm missing something, but isn't this if statement always going to be true?

carlosmondra:
Is the PR description up to date?

expertise/execute_pipeline.py
melisabok: same question posted above

expertise/models/multifacet_recommender/specter.py
```diff
@@ -169,37 +172,57 @@ def set_submissions_dataset(self, submissions_dataset):
         with open(os.path.join(self.work_dir, "specter_submission_paper_ids.txt"), 'w') as f_out:
             f_out.write('\n'.join(paper_ids_list)+'\n')
 
-    def embed_submissions(self, submissions_path=None):
-        print('Embedding submissions...')
-        metadata_file = os.path.join(self.work_dir, "specter_submission_paper_data.json")
-        ids_file = os.path.join(self.work_dir, "specter_submission_paper_ids.txt")
-
+    def _embed_papers(self, metadata_file, output_path, cached_embeddings, desc, emb_attr):
```
melisabok: can we store the cache embeddings as a property of the class instead of passing this a param everywhere?

expertise/models/specter2_scincl/predictor.py
```diff
@@ -10,16 +10,52 @@ def _build_embedding_jsonl(self, paper, embedding):
...
+            emb = cached.get(paper_id)
+            if emb is not None:
+                embeddings[paper_id] = emb
+            else:
+                remaining[paper_id] = paper
+        if cached:
+            print(f"Reusing {len(embeddings)} cached {desc.lower()} embeddings; computing {len(remaining)}.")
+
+        for batch_data in tqdm(self._fetch_batches(remaining, self.batch_size), desc=desc, total=int(len(remaining.keys()) / self.batch_size), unit="batches"):
+            for item in self._batch_predict(batch_data):
```
melisabok, line 32: can _batch_predict implementation evaluate if the paper is in the cache or not instead of computing the remaining first?

expertise/execute_expertise.py
```diff
@@ -53,15 +60,27 @@ def execute_expertise(config):
...
-        )
+            if cached_publication_embeddings is not None or cached_submission_embeddings is not None:
+                new_embeddings['pub2vec.jsonl'] = predictor.embed_publications(
+                    publications_path=publication_path,
+                    cached_publications=cached_publication_embeddings.get('specter')
+                )
+                new_embeddings['sub2vec.jsonl'] = predictor.embed_submissions(
+                    submissions_path=submissions_path,
+                    cached_submissions=cached_submission_embeddings.get('specter')
+                )
```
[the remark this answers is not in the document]
cagostino: Fixed by consolidating to a single `cached_embeddings` parameter. All `.get()` calls now use `(cached_embeddings or {}).get('specter')` (and `'scincl'`), so a `None` dict is normalized to `{}` before any lookup.

expertise/execute_pipeline.py
```diff
@@ -197,8 +187,118 @@ def run_pipeline(
...
+                    if count > 0:
+                        print(f"Pre-populated {count} embeddings from global cache ({cache_key})", flush=True)
+                print(f"Global cache lookup completed ({total_cached}/{len(note_ids)} embeddings found)", flush=True)
+            else:
+                print("No note IDs found; skipping global cache lookup", flush=True)
+                cached_publication_embeddings = {}
+                cached_submission_embeddings = {}
+        except Exception as e:
+            print(f"Global cache lookup failed: {e}", flush=True)
+            raise
```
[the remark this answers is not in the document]
cagostino: the cache lookup is now wrapped in `try/except`. On failure it prints a warning and falls back to empty cached embeddings, so the pipeline continues.

expertise/models/specter2_scincl/specter.py
cagostino: I've extracted it as a method to the base Predictor class that the models instantiate from

expertise/create_dataset.py
cagostino: cache migrated so embedding_date uses epoch timestamp as well and mdate now no longer converted

expertise/execute_expertise.py
cagostino: the two have been consolidated, there are still separate methods for the two to be embedded that have the paths hardcoded for loading the relevant json, re-factoring that as well to simplify

expertise/execute_pipeline.py
cagostino: in my tests so far, it's been usually somewhere between 40 seconds and two minutes depending on how many results there are to write

expertise/models/multifacet_recommender/specter.py
cagostino: adding this in

expertise/models/multifacet_recommender/specter.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, specter_dir, work_dir, average_score=False, max_score=True, b
         else:
             self.cuda_device = torch.device("cpu")
         self.preliminary_scores = None
+        self.publication_embeddings = {}
+        self.submission_embeddings = {}
+        self.cached_embeddings = {}
```
carlosmondra, line 68: I think this needs to be updated too

carlosmondra: Nevermind. I think I understand.

expertise/models/specter2_scincl/specter.py
```diff
@@ -252,6 +215,28 @@ def load_emb_file(emb_file, paper_id_to_weight=None):
...
+                    bad_id_set.add(paper_id)
+                if paper_id_to_weight is not None:
+                    weight_list.append(paper_id_to_weight.get(paper_id, 1.0))
+                id_list.append(paper_id)
+                emb_list.append(paper_emb)
+            emb_tensor = torch.tensor(emb_list, device=torch.device('cpu'))
+            emb_tensor = emb_tensor / (emb_tensor.norm(dim=1, keepdim=True) + 0.000000000001)
+            weight_tensor = torch.tensor(weight_list, device=torch.device('cpu'), dtype=torch.float32)
+            print(len(bad_id_set))
+            return emb_tensor, id_list, bad_id_set, weight_tensor
```
carlosmondra: This is also repeated code. If we can avoid it, it would be ideal

cagostino: abstracted now into predictor class

expertise/models/specter2_scincl/predictor.py
cagostino: i think it makes more sense to keep it before because if we send in a batch of 16 then if there are cache hits we'd send an incomplete batch. or do you mean that the partitioning of the batches should be done inside the batch predict?

expertise/execute_expertise.py
cagostino: this has been removed, now we set the cached embeddings on the predictor (execute_expertise.py:66 (https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/execute_expertise.py#L66))  and embed handles the empty-dict case  (predictor.py:15-33 (https://github.com/openreview/openreview-expertise/blob/feat/embedding-cache/expertise/models/specter2_scincl/predictor.py#L15-L33))

expertise/models/specter2_scincl/predictor.py
melisabok: Discussed in a video call and decided it to leave it as it is

melisabok (approved):

setup.py
```diff
@@ -35,7 +35,8 @@
         'importlib-metadata==4.13.0',
         'google-cloud-storage',
         'google-cloud-aiplatform',
-        'bullmq==2.11.0'
+        'bullmq==2.11.0',
+        'pyarrow'
```
carlosmondra: Can you pin the version that you will need? Also, can you do the same for the other libraries in this file?

carlosmondra (approved):

PR #397, opened 2026-07-21
Add signed URL endpoint for cloud job results

cagostino:
Adds GET /expertise/results/all to return a signed GCS URL for a cloud job's results file.

- Same auth as /expertise/results (job owner or superuser).
- Cloud only.
- Returns a 302 redirect to a V4 signed URL expiring in 5 minutes.
- Default returns full results; pass sparse=true for the sparse score file.
- Full result selection prefers scores.csv / scores.jsonl, falling back to the .pt matrix.
- Uses service account impersonation of GCP_URL_SIGNER_SERVICE_ACCOUNT because the runtime GCE SA cannot sign URLs directly.


expertise/service/utils.py
```diff
@@ -1528,6 +1532,81 @@ def stream_score_file(blob):
...
+        if len(sparse_score_files) == 1:
+            target_blob = sparse_score_files[0]
+        else:
+            non_sparse_score_files = [blob for blob in score_files if 'sparse' not in blob.name]
+            if len(non_sparse_score_files) != 1:
+                raise openreview.OpenReviewException(
+                    f'Internal Error: incorrect non-sparse score files found expected [1] found {len(non_sparse_score_files)}'
+                )
+            target_blob = non_sparse_score_files[0]
+
```
carlosmondra: Correct me if I'm wrong, but I think this is repeated code.
Also, why are you returning sparse scores as default? I think that at least the default should be the full matrix or provide an option for the user

expertise/service/routes.py
```diff
@@ -418,6 +418,64 @@ def metadata():
...
+        if not flask.current_app.config.get('USE_GCP'):
+            raise openreview.OpenReviewException('Bad request: signed URLs are only available in cloud mode')
+
+        duration = flask.request.args.get('duration', '10')
+        try:
+            duration = int(duration)
+        except ValueError:
+            raise openreview.OpenReviewException('Bad request: duration must be an integer number of minutes')
+        if duration <= 0 or duration > 1440:
+            raise openreview.OpenReviewException('Bad request: duration must be between 1 and 1440 minutes')
```
carlosmondra: I don't think we should allow the user to select the duration of the signed url validity, but even if you think otherwise, 1440 min as the upper limit seems excesive.

expertise/service/routes.py
```diff
@@ -418,6 +418,55 @@ def metadata():
         flask.current_app.logger.error(str(error_handle), exc_info=True)
         return flask.jsonify(format_error(500, 'Internal server error: {}'.format(error_handle))), 500
 
+@BLUEPRINT.route('/expertise/signed-url', methods=['GET'])
```
melisabok: we need a more meaningful name for this endpoint that represents that the user is requesting to get all the scores

carlosmondra: Btw, do we want this endpoint to only return the full score matrix or should we allow the user to also be able to download the sparse scores?

expertise/service/routes.py
```diff
@@ -418,6 +418,62 @@ def metadata():
...
+        if not flask.current_app.config.get('USE_GCP'):
+            raise openreview.OpenReviewException('Bad request: signed URLs are only available in cloud mode')
+
+        fmt = flask.request.args.get('format', 'full')
+        if fmt not in ('full', 'sparse'):
+            raise openreview.OpenReviewException("Bad request: format must be 'full' or 'sparse'")
+
+        expertise_service = get_expertise_service(flask.current_app.config, flask.current_app.logger)
+        result = expertise_service.get_expertise_signed_url(user_id, job_id, fmt)
+        return flask.jsonify({'signedUrl': result}), 200
```
carlosmondra: Is it possible to do something like a redirect so that the file is downloaded directly using the signed url instead of providing the signed url?

expertise/service/routes.py
```diff
@@ -418,6 +418,62 @@ def metadata():
...
+    try:
+        user_id = g.user_id
+        flask.current_app.logger.debug('GET receives ' + str(flask.request.args))
+        job_id = flask.request.args.get('jobId', None)
+        if job_id is None or len(job_id) == 0:
+            raise openreview.OpenReviewException('Bad request: jobId is required')
+        if not flask.current_app.config.get('USE_GCP'):
+            raise openreview.OpenReviewException('Bad request: signed URLs are only available in cloud mode')
+
+        fmt = flask.request.args.get('format', 'full')
```
carlosmondra: what does `fmt` stand for?

cagostino: output format being sparse versus full

expertise/service/routes.py
cagostino: looking into it

expertise/service/routes.py
melisabok: be careful because we already use format for json vs csv :|

expertise/service/routes.py
cagostino: have adopted '/expertise/results/all'

expertise/service/routes.py
cagostino: I'll change it to be sparse=False default

.github/workflows/deploy-dev.yml
```diff
@@ -138,5 +138,9 @@ jobs:
 
           for i in ${!instance_names[@]}; do
             echo Deploying to ${instance_names[$i]}
-            gcloud compute ssh --zone ${zones[$i]} --tunnel-through-iap openreview@${instance_names[$i]} --command "bash /home/openreview/bin/deploy-expertise.sh $BRANCH" --quiet
+            if [[ "${instance_names[$i]}" == "openreview-dev-matcher" ]]; then
+              gcloud compute ssh --zone ${zones[$i]} --tunnel-through-iap openreview@${instance_names[$i]} --command "bash /home/openreview/dev/openreview-expertise/expertise-dev.sh $BRANCH" --quiet
+            else
+              gcloud compute ssh --zone ${zones[$i]} --tunnel-through-iap openreview@${instance_names[$i]} --command "bash /home/openreview/bin/deploy-expertise.sh $BRANCH" --quiet
+            fi
```
carlosmondra: I thought that the dev site had the openreview expertise deployment

cagostino: it should be the case but the deploy still points to this instance and failed when I tried the first time so was trying to do this to get past it for time being https://github.com/openreview/openreview-expertise/actions/runs/30025901996/job/89270171205#step:14:520

carlosmondra: As discussed, let's undo this. Tyler is going to change the name of the instance and we are not supposed to deploy to openreview-dev-expertise anyway.

tests/test_expertise_cloud_service.py
carlosmondra, line 1: If we are only testing that the signed url can be used to download files from the bucket, wouldn't it be better to create a fake job in the test bucket and download it instead of mocking everything in the test?

cagostino: if we do this we'd have to use a real service account right?

carlosmondra (approved):

tests/test_expertise_cloud_service.py
carlosmondra: Yes, good point :)

melisabok (approved):

expertise/service/routes.py
```diff
@@ -418,6 +418,60 @@ def metadata():
...
+        job_id = flask.request.args.get('jobId', None)
+        if job_id is None or len(job_id) == 0:
+            raise openreview.OpenReviewException('Bad request: jobId is required')
+        if not flask.current_app.config.get('USE_GCP'):
+            raise openreview.OpenReviewException('Bad request: signed URLs are only available in cloud mode')
+
+        sparse = flask.request.args.get('sparse', 'false').lower() in ('true', '1', 'yes')
+
+        expertise_service = get_expertise_service(flask.current_app.config, flask.current_app.logger)
+        result = expertise_service.get_expertise_signed_url(user_id, job_id, sparse=sparse)
```
tyler-dominguez, line 453: Is there any logging of who signed_urls are for with user_id/job_id?

cagostino: not at the moment but can add

tyler-dominguez: GCP audits will just so url_signer impersonation so no insights. Not sure how important it is.

tyler-dominguez (approved):

